PR #16, opened 2015-02-09
Revert "Merge pull request #14 from Shopify/bankers_rounding"

codyfauser:
This reverts commit 61de47c004d7e4704731d85fae5fabb23ec86dc4, reversing
changes made to 8c24f677185303eaef119222040233d6b1fd0141.

@boourns @jeromecornet @Smcchoi @mikeygagnon 

Banker's rounding has no place at all in the money library. I was going to leave the option in there and change the default, but it really shouldn't be a feature at all. Shopify will need to stop using the `:banker` option before we merge this in.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -142,6 +142,10 @@ class Money
     Money.new(value.floor)
   end
 
+  def round(ndigits=0)
+    Money.new(value.round(ndigits))
+  end
+
   def fraction(rate)
     raise ArgumentError, "rate should be positive" if rate < 0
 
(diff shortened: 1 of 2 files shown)


Smcchoi:
:+1:

codyfauser:
@boourns yeah, this was the wrong fix. We should have fixed POS.

joshpc:
Yeah we discussed in person earlier -- banker's rounding was a bad call on my behalf. The research that I had done before led towards banker's rounding being the standard, but it was not.

boourns:
:+1: cool just wanted to make sure we had a plan

PR #17, opened 2015-02-23
Update RSpec to 3.2.0

sgnr:
This allows to use the newer syntax such as `expect(foo).to eq(bar)` as well as a better runner.

r: @gmalette @davidcornu


gmalette:
YAY!

davidcornu:
:+1:

PR #18, opened 2015-02-23
Money.from_cents works with amounts larger than 2**53.

sgnr:
I guess this is kind of a corner case, but why take the risk...

r: @gmalette @davidcornu @boourns

### DIFF
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -175,6 +175,11 @@ describe "Money" do
     Money.new(16.51).round(1).should == Money.new(16.5)
   end
 
+  it "can be converted from an amount of cents that cannot be represented as an IEEE 754 double precision float" do
+    amount = 2**53 + 1
+    expect(Money.from_cents(amount).cents).to eq(amount)
+  end
+
   describe "frozen with amount of $1" do
     before(:each) do
       @money = Money.new(1.00).freeze
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -98,7 +98,7 @@ class Money
   end
 
   def self.from_cents(cents)
-    Money.new(cents.round.to_f / 100)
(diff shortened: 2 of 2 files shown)


spec/money_spec.rb
```diff
@@ -175,6 +175,11 @@
     Money.new(16.51).round(1).should == Money.new(16.5)
   end
 
+  it "can be converted form an amount of cents that cannot be represented as an IEEE 754 double precision float" do
```
davidcornu: form &rarr; from

sgnr: fixed

davidcornu:
:+1:

spec/money_spec.rb
```diff
@@ -175,6 +175,11 @@
...
+    amount = 2**53 + 1
+    expect(Money.from_cents(amount).cents).to eq(amount)
```
gmalette: Because many methods use `@value`, the cents are good, but the `.to_s`, `to_json`, `+`, etc, will produce bad results

gmalette:
```
> m.cents
=> 9007199254740993
irb(main):037:0> m.to_s
=> "90071992547409.94"
```

sgnr: This PR is about `Money.from_cents`. I'll open another one for `to_s`.

gmalette: sure

gmalette: but then what's the point if this PR doesn't have correct values either?

sgnr: `Money#value` and `Money#cents` is fixed.

sgnr: It's just `to_s` that still has to be fixed.

sgnr:
`to_s` is pretty tricky to get right and I don't want to risk a bug in Shopify.

PR #22, opened 2015-10-07
Add a non-ActiveRecord money accessor

tjoyal:
Proposing a new module to support the equivalent of `MoneyColumn::ActiveRecordHooks` for non `ActiveRecord::Base` objects.

```ruby
class NormalObject
  include MoneyAccessor
  
  money_accessor :price

  def initialize(price)
    @price = price
  end
end

object = NormalObject.new(12.34)
object.price # => #<Money value:12.34>
```

The code is somewhat similar to `MoneyColumn::ActiveRecordHooks`, but I believe there was not enough in common to abstract what would be shared. I think it would simply complexify things without any real gain.

### DIFF
--- a/lib/money_accessor.rb
+++ b/lib/money_accessor.rb
@@ -1,0 +1,27 @@
+module MoneyAccessor
+  def self.included(base)
+    base.extend(ClassMethods)
+  end
+
+  module ClassMethods
+    def money_accessor(*columns)
+      Array(columns).flatten.each do |name|
+        define_method(name) do
+          value = instance_variable_get("@#{name}")
+          value.blank? ? nil : Money.new(value)
+        end
+
+        define_method("#{name}=") do |value|
+          if value.blank? || !value.respond_to?(:to_money)
+            instance_variable_set("@#{name}", nil)
+            nil
+          else
+            money = value.to_money
+            instance_variable_set("@#{name}", money.value)
+            money
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -2,14 +2,14 @@ module MoneyColumn
   module ActiveRecordHooks
     def self.included(base)
       base.extend(ClassMethods)
-    end                   
+    end
 
     module ClassMethods
       def money_column(*columns)
-        [columns].flatten.each do |name|
+        Array(columns).flatten.each do |name|
           define_method(name) do
             value = read_attribute(name)
-            value.blank? ? nil : Money.new(read_attribute(name))
+            value.blank? ? nil : Money.new(value)
           end
 
           define_method("#{name}_before_type_cast") do
@@ -31,4 +31,3 @@ module MoneyColumn
     end
   end
 end
-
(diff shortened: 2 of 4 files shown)


tjoyal:
@davidcornu @sgnr Could I have your opinions and review on this please.

spec/money_accessor_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+describe NormalObject do
+  before(:each) do 
+    @object = Object.new
+    @money = Money.new("1.00")
+  end
+
+  it "generate an attribute reader that returns a money object" do
+    object = NormalObject.new(100)
+
+    object.price.should == Money.new(100)
```
sgnr: `expect(object.price).to eq(Money.new(100))`

lib/money_accessor.rb
```diff
@@ -0,0 +1,27 @@
...
+  module ClassMethods
+    def money_accessor(*columns)
+      [columns].flatten.each do |name|
+        define_method(name) do
+          value = instance_variable_get("@#{name}")
+          value.blank? ? nil : Money.new(value)
+        end
+
+        define_method("#{name}=") do |value|
+          if value.blank? || !value.respond_to?(:to_money)
```
sgnr: I don't like that it's just discarding it, but I guess the point is to have the exact same behaviour as `ActiveRecordHooks`.

spec/money_accessor_spec.rb
tjoyal: Done, got rid of the legacy should syntax

lib/money_accessor.rb
tjoyal: Yep exactly what I had in mind. I prefer everything working the same way.

sgnr:
I think that makes sense. :shipit:

lib/money_accessor.rb
```diff
@@ -0,0 +1,27 @@
+module MoneyAccessor
+  def self.included(base)
+    base.extend(ClassMethods)
+  end
+
+  module ClassMethods
+    def money_accessor(*columns)
+      [columns].flatten.each do |name|
```
davidcornu: `Array(columns).each`?

lib/money_column/active_record_hooks.rb
```diff
@@ -2,14 +2,14 @@ module MoneyColumn
   module ActiveRecordHooks
     def self.included(base)
       base.extend(ClassMethods)
-    end
```
davidcornu: facepalm

spec/money_accessor_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+end
+
+describe NormalObject do
+  before(:each) do 
+    @object = Object.new
+    @money = Money.new("1.00")
+  end
+
+  it "generate an attribute reader that returns a money object" do
```
davidcornu: generate**s**

davidcornu:
Minor stuff. LGTM

tjoyal:
@davidcornu All fixed and :green_apple: !

davidcornu:
:shipit:

PR #27, opened 2015-10-27
Use struct accessors for storing/retrieving the money value

maximebedard:
When deserializing a struct, the variable information are stored in an internal hash rather than ivars. This PR uses `:[]` on the `Struct` to get and set the correct value.

@sgnr @tjoyal

### DIFF
--- a/lib/money_accessor.rb
+++ b/lib/money_accessor.rb
@@ -5,43 +5,28 @@ module MoneyAccessor
 
   module ClassMethods
     def money_accessor(*columns)
+      variable_get = self <= Struct ? :[]  : :instance_variable_get
+      variable_set = self <= Struct ? :[]= : :instance_variable_set
+
       Array(columns).flatten.each do |name|
+        variable_name = self <= Struct ? name : "@#{name}"
+
         define_method(name) do
-          value = _variable_get(name)
+          value = public_send(variable_get, variable_name)
           value.blank? ? nil : Money.new(value)
         end
 
         define_method("#{name}=") do |value|
           if value.blank? || !value.respond_to?(:to_money)
-            _variable_set(name, nil)
+            public_send(variable_set, variable_name, nil)
             nil
           else
             money = value.to_money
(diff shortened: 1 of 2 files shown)


lib/money_accessor.rb
```diff
@@ -7,21 +7,41 @@ module ClassMethods
     def money_accessor(*columns)
       Array(columns).flatten.each do |name|
         define_method(name) do
-          value = instance_variable_get("@#{name}")
+          value = _variable_get(name)
```
sgnr: Not sure it matters, but could move the `case self` outside the `define_method` to avoid a dynamic check every time the accessor is called.

sgnr:
That must have been fun to debug lol

tjoyal:
Not sure if this was a good idea in the first place (`money_accessor`) but this is sure the logical next step 

Let's see if you can make @sgnr happy then :+1: for me.

sgnr:
> Let's see if you can make @sgnr happy then :+1: for me.

lol you can bring me food, I'm not difficult.

maximebedard:
I think i've been able to extract the type check correctly, what do you think @sgnr?

lib/money_accessor.rb
```diff
@@ -5,19 +5,24 @@ def self.included(base)
 
   module ClassMethods
     def money_accessor(*columns)
+      variable_get = self <= Struct ? :[]  : :instance_variable_get
+      variable_set = self <= Struct ? :[]= : :instance_variable_set
```
sgnr: I know it's unlikely, but what if we include that in a class that defines `:<=`? Can you use `is_a?` instead? It's also more readable imo.

sgnr: Ah wait `self` is the `Class`.

sgnr:
:raised_hands:

maximebedard:
@tjoyal do I have your blessing?

tjoyal:
Are we writing javascript here? :green_heart:

PR #28, opened 2015-11-17
Allow sum of split in #allocate to be > 1.

sgnr:
I need to split money in a line item discount, so I wanted to do something line `amount.split([3, 4])`, but that's not allowed. There seem to be no reason for this? The git history doesn't go that far ¯\_(ツ)_/¯.

r: @tjoyal @maximebedard

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -170,7 +170,6 @@ class Money
   #   Money.new(100, "USD").allocate([0.33,0.33,0.33]) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def allocate(splits)
     allocations = splits.inject(0.0) {|sum, i| sum += i }
-    raise ArgumentError, "splits add to more than 100%" if allocations > 1.0
 
     left_over = cents
 
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -313,8 +313,8 @@ describe "Money" do
       expect(moneys[2].cents).to eq(33)
     end
 
-    specify "#allocate requires total to be less then 1" do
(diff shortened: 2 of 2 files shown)


tjoyal:
Could that be in any way related to float accuracy issue? One other reason could be to ensure all the weights are floats (as opposed to int) so you do not have divisions of `int / int` that would behave differently than `(int / float).floor` ?

sgnr:
Since `inject` has an initial value of `0.0`, it's result will also be a float.

sgnr:
So we never get the `int / int` case.

tjoyal:
I don't see why prevent this. As it is $ we might want to dig deeper, but for me :+1:

sgnr:
Hum [Ruby money does seem to disallow allocations > 1](https://github.com/RubyMoney/money/blob/e5783c8078a8fb9e18e9cfd1b56f61265bbe072e/lib/money/money.rb#L490..L492). But I can't find any justification for that in the comments or issues... in their example they explicitly show and example where the sum is < 1, so it _has_ to normalize? ¯\_(ツ)_/¯

sgnr:
@jeromecornet, @gmalette what do you think of this? It is safe, right?

gmalette:
I don't understand how this helps

jeromecornet:
IMO that kind of breaks the semantics of `allocate` even if the math would work.
What prevents you from normalizing the ratios in the caller ?

sgnr:
> What prevents you from normalizing the ratios in the caller ?

Yeah, I think I'll just do that instead.

PR #29, opened 2016-02-09
Don't coerce Rationals to floats in allocate

jahfer:
Right now, if we attempt to pass `Rational` numbers to `allocate`, it will coerce those values into floats, sum them, and then raise if the resulting value is `> 1`. This is problematic for values who's rational representation sums to `1`, but when converted to float, the precision is lost and the value grows.

All of this code was pulled out of [RubyMoney/money#allocate](https://github.com/RubyMoney/money/blob/e8281aafd40af5d9b6d06cdc0d47703eb7d8cc34/lib/money/money.rb#L487-L501), and massaged to work with what we had already.

Running the new test before updating `#allocate` results in:
```
Failures:

  1) Money allocation #allocate will use rationals if provided
     Failure/Error: expect(Money.new(2.25).allocate(splits)).to eq([Money.new(1.42), Money.new(0.23), Money.new(0.16), Money.new(0.16), Money.new(0.28)])
     ArgumentError:
       splits add to more than 100%
     # ./lib/money/money.rb:173:in `allocate'
     # ./spec/money_spec.rb:322:in `block (3 levels) in <top (required)>'
```

@Shopify/merchant-data @tjoyal @jduff for review

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -169,20 +169,17 @@ class Money
   #   Money.new(5, "USD").allocate([0.3,0.7)) #=> [Money.new(2), Money.new(3)]
   #   Money.new(100, "USD").allocate([0.33,0.33,0.33]) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def allocate(splits)
-    allocations = splits.inject(0.0) {|sum, i| sum += i }
-    raise ArgumentError, "splits add to more than 100%" if allocations > 1.0
+    allocations = splits.inject(0) { |sum, n| sum + value_to_decimal(n) }
 
-    left_over = cents
-
-    amounts = splits.collect do |ratio|
-      fraction = (cents * ratio / allocations).floor
-      left_over -= fraction
-      fraction
+    if (allocations - BigDecimal("1")) > Float::EPSILON
+      raise ArgumentError, "splits add to more than 100%"
     end
 
-    left_over.times { |i| amounts[i % amounts.length] += 1 }
+    amounts, left_over = amounts_from_splits(allocations, splits)
 
-    return amounts.collect { |cents| Money.from_cents(cents) }
+    left_over.to_i.times { |i| amounts[i % amounts.length] += 1 }
+
+    amounts.collect { |cents| Money.from_cents(cents) }
   end
 
   # Split money amongst parties evenly without losing pennies.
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -169,20 +169,17 @@ def fraction(rate)
...
-    raise ArgumentError, "splits add to more than 100%" if allocations > 1.0
+    allocations = splits.inject(0) { |sum, n| sum + value_to_decimal(n) }
 
-    left_over = cents
-
-    amounts = splits.collect do |ratio|
-      fraction = (cents * ratio / allocations).floor
-      left_over -= fraction
-      fraction
+    if (allocations - BigDecimal("1")) > Float::EPSILON
```
nick-mcdonald: :+1:

nick-mcdonald:
LGTM

lib/money/money.rb
```diff
@@ -209,17 +206,24 @@ def split(num)
...
-    # subclasses of BigDecimal will be handled
-    # in the else clause
-    if value.class == BigDecimal
-      value
-    elsif value.respond_to?(:to_d)
-      value.to_d
+
+  def value_to_decimal(num)
+    if num.respond_to?(:to_d)
+      num.is_a?(Rational) ? num.to_d(16) : num.to_d
```
jeromecornet: What is the default precision for `Rational#to_d` ?

spec/money_spec.rb
```diff
@@ -316,6 +316,11 @@
     specify "#allocate requires total to be less then 1" do
       expect { Money.new(0.05).allocate([0.5,0.6]) }.to raise_error(ArgumentError)
     end
+
+    specify "#allocate will use rationals if provided" do
+      splits = [128400,20439,14589,14589,25936].map{ |num| Rational(num, 203953) } # sums to > 1 if converted to float
```
jeromecornet: :heart: the test values :laughing:

jeromecornet:
:shipit:

lib/money/money.rb
jahfer: `Rational#to_d` actually requires a precision to be passed and doesn't provide a default.

jeromecornet: ah that makes sense then :+1:

PR #32, opened 2016-07-12
Add xml serilization methods

cplehm:
Allow Money objects to be serialized as xml.

@vignesh-vs-in @nick-mcdonald

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   ruby:
-    version: ruby-2.1.1
+    version: 2.3.0
 
 database:
   override:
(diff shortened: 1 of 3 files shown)


jahfer:
Missing a test for this behaviour. Logic makes sense to me 👍

@stephenminded @maximebedard as well if you have opinions

stephenminded:
LGTM

cplehm:
Thanks! Added tests and bumped ruby version to fix bundle install issue in ci. @jahfer

vignesh-vs-in:
👍

PR #33, opened 2016-07-14
Create xml in to_xml if options present

cplehm:
to_xml should return formatted xml if options is present. Worked on in conjunction with @vignesh-vs-in. 

Please review, @jahfer @stephenminded

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -138,8 +138,8 @@ class Money
   end
 
   def to_xml(options = {})
-    if options.try(:[], :builder)
-      options[:builder].tag!(options.fetch(:root), to_s, type: "decimal")
+    if options.key?(:builder) && options.key?(:root)
+      options[:builder].tag!(options[:root], to_s, type: "decimal")
     else
       to_s
     end
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -138,11 +138,11 @@ class Money
   end
 
   def to_xml(options = {})
-    to_s
-  end
-
-  def as_xml(*args)
-    to_s
+    if options.try(:[], :builder)
+      options[:builder].tag!(options.fetch(:root), to_s, type: "decimal")
+    else
+      to_s
+    end
   end
 
   def abs
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -1,4 +1,5 @@
 require File.expand_path(File.dirname(__FILE__) + '/spec_helper')
+require 'builder'
 
 describe "Money" do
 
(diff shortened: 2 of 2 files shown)


spec/money_spec.rb
```diff
@@ -157,8 +158,14 @@
...
   end
 
-  it "returns cents in as_xml" do
-    expect(Money.new(1.00).as_xml).to eq("1.00")
+  it "returns xml in to_xml with options present" do
+    options = {  
+      :root=>"price",
+      :format=>"xml",
+      :builder=>Builder::XmlMarkup.new,
+      :skip_instruct=>true
```
stephenminded: Can you use the new Ruby hash syntax throughout this PR? IE: `skip_instruct: true`.

lib/money/money.rb
```diff
@@ -138,11 +138,11 @@ def as_json(*args)
   end
 
   def to_xml(options = {})
-    to_s
-  end
-
-  def as_xml(*args)
-    to_s
+    if options.try(:[], :builder)
+      options[:builder].tag!(options[:root], to_s, {:type=>"decimal"})
```
stephenminded: `options[:root]` is required here (unless you want to provide a default, but I can't think of a sane way to do that). So, can you use `options.fetch(:root)` which will raise a `KeyError` if it isn't provided.

vignesh-vs-in:
Fixed the ruby hashing format. Can we get a go ahead for this, if everything else is looking fine?

lib/money/money.rb
```diff
@@ -138,11 +138,11 @@ def as_json(*args)
   end
 
   def to_xml(options = {})
-    to_s
-  end
-
-  def as_xml(*args)
-    to_s
+    if options.try(:[], :builder)
+      options[:builder].tag!(options.fetch(:root), to_s, {:type=>"decimal"})
```
stephenminded: Use the new hash syntax here as well `type: "decimal"` (also, no `{}` needed)

stephenminded:
One comment, after that 🚢

lib/money/money.rb
```diff
@@ -138,11 +138,11 @@ def as_json(*args)
   end
 
   def to_xml(options = {})
-    to_s
-  end
-
-  def as_xml(*args)
-    to_s
+    if options.try(:[], :builder)
```
jahfer: `if options.key?(:builder)`

jahfer:
don't know much about the XML path of things, but LGTM 👍

jeromecornet:
Am I reading the code correctly in that it is serializing every money object under a <price> tag now?

stephenminded:
> Am I reading the code correctly in that it is serializing every money object under a <price> tag now?

If the `builder` and `root` options are provided, it will serialize the money object under the tag specified by `root`. Otherwise we'll serialize the money value as a strait up string with no tag. I'm unsure if rails provides these options by default when serializing to XML, however.

vignesh-vs-in:
Rails expects a serialized xml tag when it provides the serialization options. It does provide the options when serializing a model.

PR #35, opened 2016-10-04
raise type error if other incompatible

richardmonette:
Fixes https://github.com/Shopify/shopify/issues/84192

When there are searches like `{% assign sorted_result = search.results|sort:'price' %}` we wind up at https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L128

```
def sort(input, property = nil)
  ary = InputIterator.new(input)
  if property.nil?
    ary.sort
  elsif ary.empty? # The next two cases assume a non-empty array.
    []
  elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
    ary.sort { |a, b| a[property] <=> b[property] }
  end
end
```

Consider a search where the results are a `ProductDrop` and an `ArticleDrop`. The `ProductDrop` will respond to `price` but the `ArticleDrop` will not. When we run the overloaded operator https://github.com/Shopify/money/blob/master/lib/money/money.rb#L22 we'll try `to_money` and create a `NoMethodError`. 

This doesn't provide any particular information about the error. In Ruby, a much better error is returned

```
irb(main):001:0> 1 + nil
TypeError: nil can't be coerced into Fixnum
```

Let's do the same!

@rafaelfranca @stephenminded cc @Thibaut

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -19,14 +19,17 @@ class Money
   end
 
   def <=>(other)
+    raise TypeError, "#{other.class.name} can't be coerced into Money" unless other.respond_to?(:to_money)
     cents <=> other.to_money.cents
   end
 
   def +(other)
+    raise TypeError, "#{other.class.name} can't be coerced into Money" unless other.respond_to?(:to_money)
     Money.new(value + other.to_money.value)
   end
 
   def -(other)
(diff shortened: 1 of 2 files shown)


Thibaut:
👍

richardmonette:
I'll look into bumping Shopify tomorrow.

PR #36, opened 2016-10-11
Revert "Create xml in to_xml if options present"

richardmonette:
Reverts Shopify/money#33

This change, which apparently we don't need in core, is preventing the gem from being updated. Let's back off on this change until there is a clear need.

@vignesh-vs-in @cplehm cc @stephenminded


vignesh-vs-in:
👍

cplehm:
👍

PR #37, opened 2016-10-12
Revert "Add xml serilization methods"

richardmonette:
Reverts Shopify/money#32

https://buildkite.com/shopify/shopify-branches/builds/172534#2b32f6da-bd13-4ae6-bdac-e31c808b4cc1

^ @vignesh-vs-in @cplehm cc @stephenminded still getting a bunch of failures, I think we need to revert this too, which then gets us back to the last version which could be cleanly used with Shopify core

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   ruby:
-    version: ruby-2.1.1
+    version: 2.3.0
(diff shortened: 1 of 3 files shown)


cplehm:
👍

PR #38, opened 2016-11-17
introducing a new weighted allocation method that splits money to the maximum provided

tskorupa:
This PR aims to give greater control to the caller of `#allocate` in that a maximum allocation amount can be specified. This affects the way the remainder is assigned.

@maximebedard @rafaelfranca for review

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -189,6 +189,50 @@ class Money
     amounts.collect { |cents| Money.from_cents(cents) }
   end
 
+  # Allocates money between different parties up to the maximum amounts specified.
+  # Left over pennies will be assigned round-robin up to the maximum specified.
+  # Pennies are dropped when the maximums are attained.
+  #
+  # @example
+  #   Money.new(30.75).allocate_max_amounts([Money.new(26), Money.new(4.75)])
+  #     #=> [Money.new(26), Money.new(4.75)]
+  #
+  #   Money.new(30.75).allocate_max_amounts([Money.new(26), Money.new(4.74)]
+  #     #=> [Money.new(26), Money.new(4.74)]
+  #
+  #   Money.new(30).allocate_max_amounts([Money.new(15), Money.new(15)]
+  #     #=> [Money.new(15), Money.new(15)]
+  #
+  #   Money.new(1).allocate_max_amounts([Money.new(33), Money.new(33), Money.new(33)])
+  #     #=> [Money.new(0.34), Money.new(0.33), Money.new(0.33)]
+  #
+  #   Money.new(100).allocate_max_amounts([Money.new(5), Money.new(2)])
+  #     #=> [Money.new(5), Money.new(2)]
+  def allocate_max_amounts(maximums)
+    cents_maximums = maximums.map { |max_amount| max_amount.to_money.cents }
+    cents_maximums_total = cents_maximums.sum
+
+    splits = cents_maximums.map do |cents_max_amount|
+      BigDecimal.new(cents_max_amount.to_s) / cents_maximums_total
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -189,6 +189,52 @@ def allocate(splits)
...
+
+    while(left_over > 0) do
+      cents_amounts.zip(cents_maxima).each.with_index do |(amount, maximum), index|
+        if amount < maximum
+          cents_amounts[index] += 1
+          left_over -= 1
+        end
+        break if left_over <= 0
+      end
+      break if cents_amounts.zip(cents_maxima).none? { |(amount, maximum)| amount < maximum }
```
rafaelfranca: Can we extract this code to a method that make clear what it does?

lib/money/money.rb
```diff
@@ -189,6 +189,52 @@ def allocate(splits)
...
+    end
+
+    cents_amounts.collect { |cents| Money.from_cents(cents) }
```
rafaelfranca: This is repeated from the line 220. So maybe we can simplify this code extracting the while loop to a method and only calling it if `left_over` is not zero.

lib/money/money.rb
```diff
@@ -189,6 +189,52 @@ def allocate(splits)
...
+  #
+  #   Money.new(30).allocate_amounts([Money.new(15), Money.new(15)]
+  #     #=> [Money.new(15), Money.new(15)]
+  #
+  #   Money.new(1).allocate_amounts([Money.new(33), Money.new(33), Money.new(33)])
+  #     #=> [Money.new(0.34), Money.new(0.33), Money.new(0.33)]
+  #
+  #   Money.new(100).allocate_amounts([Money.new(5), Money.new(2)])
+  #     #=> [Money.new(5), Money.new(2)]
+  def allocate_amounts(maxima)
```
gmalette: - I know the latin plural of maximum is maxima, but maximums is accepted in english
- Should this just be called `allocate_discounts` or `allocate_to_max` or something? 
  - It then follows that the argument could be called `amounts`

lib/money/money.rb
```diff
@@ -189,6 +189,52 @@ def allocate(splits)
...
+    cents_total = cents_maxima.sum
+    splits = cents_maxima.map do |cents_maximum_amount|
+      BigDecimal.new(cents_maximum_amount.to_s) / cents_total
+    end
+
+    cents_amounts, left_over = amounts_from_splits(1, splits, [cents, cents_total].min)
+    if left_over.zero?
+      amounts = cents_amounts.collect { |cents| Money.from_cents(cents) }
+      return amounts
+    end
```
gmalette: I think this `if` is unnecessary

lib/money/money.rb
tskorupa: > I know the latin plural of maximum is maxima, but maximums is accepted in english

👍 



> Should this just be called allocate_discounts or allocate_to_max or something?

It has nothing to do with discounts from the Money gem perspective. I like the suggestion of `allocate_max_amounts`

tskorupa:
ping for review after addressing latest comments

lib/money/money.rb
```diff
@@ -189,6 +189,50 @@ def allocate(splits)
...
+      break unless left_over > 0
+
+      max_amount = cents_maximums[index]
+      next unless amount < max_amount
+
+      left_over -= 1
+      cents_amounts[index] += 1
+    end
+
+    cents_amounts.collect { |cents| Money.from_cents(cents) }
```
gmalette: `collect` here vs `map` up there 🚲 🏠

PR #39, opened 2016-12-07
Remove the ability to multiply a Money by a Float or BigDecimal

matthelm:
*work in progress*

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -34,7 +34,11 @@ class Money
   end
 
   def *(numeric)
-    Money.new(value * numeric)
+    if numeric.is_a?(Float)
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -34,7 +34,11 @@ def -(other)
   end
 
   def *(numeric)
-    Money.new(value * numeric)
+    if numeric.is_a?(Float)
+      Money.new(value.to_r * numeric.rationalize)
+    else
+      Money.new(value.to_r * numeric.to_r)
```
matthelm: @AnotherJoSmith Do you know of an example where this `else` case is required?

Issue #13, opened 2014-06-12
Parsing is incorrect if there are more than two numbers after the decimal point


davidcornu:
**Shopify/money**

```
irb(main):001:0> require 'money'
=> true
irb(main):002:0> "1.11111111".to_money
=> #<Money value:111111111.00>
irb(main):003:0> "1.11111111".to_money.to_f
=> 111111111.0
```

**RubyMoney/monetize** (https://github.com/RubyMoney/monetize)

```
irb(main):001:0> require 'monetize'
=> true
irb(main):002:0> "1.11111111".to_money
=> #<Money fractional:111 currency:USD>
irb(main):003:0> "1.11111111".to_money.to_f
=> 1.11
```

Discovered by @byroot

NBowes:
I have a partner who is experiencing this same issue with the `UsageCharge` API. 
[Internal Dashboard](https://app.shopify.com/services/internal/shops/18485575)

```
usage_charge = {
'price': 1.035,
'description': 'ION usage charge'
}
usage_charge = ShopifyAPI::UsageCharge.new(usage_charge)
```

They are seeing the price being charged is `1035` instead of a rounded `1.04`. They also sent in an example of what this looks like for their charges:

![](https://screenshot.click/04-12-nh4kf-t9hyb.png)

Just wondering if we've made any progress on this yet?

Thanks!
[Zendesk Ticket](https://shopify.zendesk.com/agent/tickets/6250376)

PR #43, opened 2017-01-17
Fix parsing for digits with more than 2 decimals

samirElm:
Hi, it is my first contribution to an open source project so let me know if there is other things I can do to make this change approved :)

## Problem
Issue #13 

```
irb(main):001:0> require 'money'
=> true
irb(main):002:0> "1.11111111".to_money
=> #<Money value:111111111.00>
irb(main):003:0> "1.11111111".to_money.to_f
=> 111111111.0
```

## Changes
Regex in MoneyParser now matched and substitute digits with more than 2 decimals to digits with 2 decimals => **1.1234** becomes **1.23**

### DIFF
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -5,29 +5,29 @@ class MoneyParser
   def self.parse(input)
     new.parse(input)
   end
-  
+
   def parse(input)
     Money.new(extract_money(input))
   end
-  
+
   private
   def extract_money(input)
     return ZERO_MONEY if input.to_s.empty?
-    
+
     amount = input.scan(/\-?[\d\.\,]+/).first
-            
+
     return ZERO_MONEY if amount.nil?
-    
-    # Convert 0.123 or 0,123 into what will be parsed as a decimal amount 0.12 or 0.13
-    amount.gsub!(/^(-)?(0[,.]\d\d)\d+$/, '\1\2')
-            
+
+    # Convert amount with more than 3 decimals to amount with 2 decimals
+    amount.gsub!(/^(-)?(\d{1,}[,.]\d\d)[1-9]+$/, '\1\2')
+
     segments = amount.scan(/^(.*?)(?:[\.\,](\d{1,2}))?$/).first
-        
-    return ZERO_MONEY if segments.empty?    
-    
+
+    return ZERO_MONEY if segments.empty?
+
     amount   = segments[0].gsub(/[^-\d]/, '')
     decimals = segments[1].to_s.ljust(2, '0')
-    
+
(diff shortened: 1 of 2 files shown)


pate:
Is this in production? It seems to affect Tradegecko integration.

samirElm:
No, the PR has not been merged yet.

lib/money/money_parser.rb
```diff
@@ -5,29 +5,29 @@ class MoneyParser
...
-            
+
     return ZERO_MONEY if amount.nil?
-    
-    # Convert 0.123 or 0,123 into what will be parsed as a decimal amount 0.12 or 0.13
-    amount.gsub!(/^(-)?(0[,.]\d\d)\d+$/, '\1\2')
-            
+
+    # Convert amount with more than 3 decimals to amount with 2 decimals
+    amount.gsub!(/^(-)?(\d{1,}[,.]\d\d)[1-9]+$/, '\1\2')
```
pate: Why complect parsing and rounding?

pate:
Does Shopify support storing and display of prices with more than two decimal prices? I'm sure there are currencies where e.g. 2.1234 is common?

We need prices with 3+ decimal points to correctly calculate tax-inclusive prices (e.g. with 14% VAT, 2999/1.14 = 2630.701754386, whereas 2630.70 * 1.14 = 2998.998, which does not round back up to 2999.00). Tradegecko is sending our tax-exclusive prices to Shopify e.g. as 1234.997 but they are being interpreted as `1234997.00`.

samirElm:
Exactly, that is what this PR is about. Instead of interpreting `1234.997` as `1234997.0`, it will interpret it as `1234.99`. See issue #13.

And yes, there are currencies with more than 3 decimals (not so much as far as I know). I think the best way to fix this is to implement **internationalization**.

theronic:
Any update on this? I'm wary of parsing & rounding in the same line of code.

theronic:
When will this land? This is preventing us from using Tradegecko with Shopify.

elfassy:
@samirElm Thanks for this PR. `MoneyParser` will be deprecated in the near future. If you need the parsing functionality, please add it to your app
```ruby
class YourMoneyParser
 # you can copy the code from MoneyParser you have in this PR
end

Money.parser = YourMoneyParser
```

bizla:
@bdewater is it reasonable to tag this as the better alternative for https://github.com/Shopify/shopify/pull/134937? 
If so, we should get this out ASAP.

elfassy:
@bizla I've done a more robust refactoring of the money parser here: https://github.com/Shopify/money/pull/125

PR #44, opened 2017-03-07
Introduce Money.from_amount for compatibility with RubyMoney

bdewater:
…and by extension offsite_payments. 

We use this gem in Shopify instead of `money` on Rubygems. The API's were similar enough until https://github.com/activemerchant/offsite_payments/pull/217 introduced a change which broke upgrading offsite_payments in Shopify. This PR aims to restore compatibility between ShopifyMoney and RubyMoney to unblock ourselves until a [better solution](https://github.com/Shopify/shopify/issues/95949) is found.

ShopifyMoney.new calls `#value_to_decimal` to either cast a Numeric object to BigDecimal, or have BigDecimal parse the input value. RubyMoney [does something similar for .from_amount](https://github.com/RubyMoney/money/blob/557561ef7259b8d2ab6a3603d0a5335e90842644/lib/money/money.rb#L216-L236).

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -6,11 +6,11 @@ class Money
 
   attr_reader :value, :cents
 
-  def self.from_amount(value)
-    new(value)
+  class << self
+    alias_method :from_amount, :new
   end
 
-  def initialize(value = 0)
+  def initialize(value = 0, _currency = nil)
     raise ArgumentError if value.respond_to?(:nan?) && value.nan?
 
     @value = value_to_decimal(value).round(2)
@@ -104,7 +104,7 @@ class Money
   end
 
   def self.empty
-    Money.new
+    Money.new(0)
   end
 
   def self.from_cents(cents)
@@ -270,8 +270,10 @@ class Money
   def value_to_decimal(num)
     if num.respond_to?(:to_d)
       num.is_a?(Rational) ? num.to_d(16) : num.to_d
-    else
+    elsif num.is_a?(Money)
       BigDecimal.new(num.to_s)
+    else
+      raise ArgumentError, "value_to_decimal could not parse #{num.inspect}"
     end
   end
 
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -6,6 +6,10 @@ class Money
 
   attr_reader :value, :cents
 
+  def self.from_amount(value)
+    new(value)
+  end
+
   def initialize(value = 0)
     raise ArgumentError if value.respond_to?(:nan?) && value.nan?
 
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -39,7 +39,7 @@ describe "Money" do
   end
 
   it "raises error if added other is not compatible" do
-    expect{ Money.new(5.00) + nil }.to raise_error
+    expect{ Money.new(5.00) + nil }.to raise_error(TypeError)
   end
 
   it "is able to add $0 + $0" do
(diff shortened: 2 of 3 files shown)


spec/money_spec.rb
```diff
@@ -485,4 +485,16 @@
...
+    it "accepts numeric values" do
+      expect(Money.from_amount(1)).to eq Money.from_cents(1_00)
+      expect(Money.from_amount(1.0)).to eq Money.from_cents(1_00)
+      expect(Money.from_amount("1".to_d)).to eq Money.from_cents(1_00)
+    end
+
+    it "raises ArgumentError with unsupported argument" do
+      expect { Money.from_amount(Object.new) }.to raise_error(ArgumentError)
+    end
+  end
```
elfassy: real money tests https://github.com/RubyMoney/money/blob/557561ef7259b8d2ab6a3603d0a5335e90842644/spec/money_spec.rb#L121-L182

lib/money/money.rb
```diff
@@ -6,6 +6,10 @@ class Money
 
   attr_reader :value, :cents
 
+  def self.from_amount(value)
+    new(value)
+  end
```
elfassy: what happens when someone passes a currency here? Example: https://github.com/activemerchant/offsite_payments/blob/71f6cdcdfaeb7a35acd5e525d8f86565234778e8/lib/offsite_payments/integrations/easy_pay.rb#L104

Krystosterone:
Mind linking to the big write-up you guys did on what's the problem we're encountering and how it solves it?

lib/money/money.rb
bdewater: That'll work because offiste_payment has this line just for us: https://github.com/activemerchant/offsite_payments/blob/e90298b1801552070ca9cee83f649363567cb8a9/lib/offsite_payments/notification.rb#L37

It's not a perf hotspot AFAIK but a conditional checking for [the method arity](https://ruby-doc.org/core-2.3.0/Method.html#method-i-arity) would be cleaner there.

spec/money_spec.rb
```diff
@@ -485,4 +485,16 @@
     end
 
   end
+
+  describe "from_amount quacks like RubyMoney" do
```
Krystosterone: 😆 😢

lib/money/money.rb
Krystosterone: Nit: Might as well 
```
class << self
  alias_method :new, :from_amount
end
```

bdewater:
Fixed all the things, ready for re-review!

lib/money/money.rb
```diff
@@ -266,8 +270,10 @@ def all_rational?(splits)
   def value_to_decimal(num)
     if num.respond_to?(:to_d)
       num.is_a?(Rational) ? num.to_d(16) : num.to_d
-    else
+    elsif num.is_a?(Money) || num.is_a?(Numeric) || num.is_a?(String)
```
Krystosterone: Where did the need for this come from?

bdewater: In order for the "raises ArgumentError with unsupported argument" spec from RubyMoney to pass we can't just call `num#to_s` and pass it to BigDecimal like we did. I couldn't think of a way to duck(type) myself out of this one 🤔

elfassy: When would we have `num.is_a?(Numeric) || num.is_a?(String)`? Both respond to the `.to_d` in theory

bdewater: Derp, you're right!

PR #49, opened 2017-04-27
Make MoneyColumn use the new AR::Type API so that the values get cached

casperisfine:
Followup: https://github.com/Shopify/money/pull/48

### Context

After some profiling, some API endpoints making heavy use of Money objects spend a big chunk of their time in the money gem:
capture d ecran 2017-04-27 a 16 03 59

### Perf investigation

It turns out instantiating a `Money` instance is quite costly. There is a lot of parsing happening even for simple known cases. I ran the following simple benchmark:

```ruby
Benchmark.ips do |x|
  x.report('BigDecimal.new(int)') { BigDecimal(0) }
  x.report('BigDecimal.new(string)') { BigDecimal('0') }
  x.report('Money.new(int)') { Money.new(0) }
  x.report('Money.new(string)') { Money.new('0') }
end
```

```
 BigDecimal.new(int)    74.779k i/100ms
BigDecimal.new(string)
                       104.490k i/100ms
      Money.new(int)    26.061k i/100ms
   Money.new(string)    28.897k i/100ms
 BigDecimal.new(int)      1.029M (± 5.0%) i/s -      5.160M in   5.028362s
BigDecimal.new(string)
                          1.588M (± 5.8%) i/s -      7.941M in   5.021061s
      Money.new(int)    297.115k (± 4.7%) i/s -      1.485M in   5.012021s
   Money.new(string)    330.913k (± 4.0%) i/s -      1.676M in   5.073410s
```

So money is adding a 5-8X overhead over just instantiating a BigDecimal.

### Money column caching

The current money column implementation re-instanciate the `Money` instance every time the column is accessed. This is incredibly wasteful given that the parsing is far from free and that the value is immutable.

This PR update the AR module using the new attributes API so that the casting only happens once per column.

@wvanbergen @DazWorrall @rafaelfranca  @bdewater @elfassy @pi3r

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -1,4 +1,15 @@
 module MoneyColumn
+  class Type < ActiveRecord::Type::Value
+    def cast(value)
+      return nil if value.blank? || !value.respond_to?(:to_money)
+      value.to_money
+    end
+
+    def serialize(money)
+      money.value
+    end
+  end
+
   module ActiveRecordHooks
     def self.included(base)
       base.extend(ClassMethods)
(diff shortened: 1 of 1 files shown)


wvanbergen:
> Nice! It will break compatibility with < 5 version of rails but I don't think anyone else is using this gem (it's not even versioned)

casperisfine:
I'd need to double check, but I think 4.2 supports this API. So it's not *that* hard of a compatibility breakage.

lib/money_column/active_record_hooks.rb
```diff
@@ -7,25 +18,7 @@ def self.included(base)
     module ClassMethods
       def money_column(*columns)
```
wvanbergen: Should we deprecate this method? Not sure if it adds much value over using the attribute API directly.

casperisfine: Yeah sure.

casperisfine: I think I'll merge as is for now, because it seems the entire gem is deprecated anyway.

PR #50, opened 2017-05-02
Cache Money.new(0) like Money.empty

casperisfine:
Followup to https://github.com/Shopify/money/pull/48

In the end this didn't make that much difference because many people use `Money.new(0)` instead of `Money.empty`. It's true that it's quite more readable.

I think we might as well cache that call too.

@wvanbergen @DazWorrall for review please.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -8,6 +8,16 @@ class Money
 
   class << self
     alias_method :from_amount, :new
+
+    def new(value = 0, _currency = nil)
+      if value == 0
+        @empty ||= from_amount(0)
+      else
+        from_amount(value, _currency)
+      end
+    end
+
+    alias_method :empty, :new
   end
 
   def initialize(value = 0, _currency = nil)
@@ -16,6 +26,7 @@ class Money
     @value = value_to_decimal(value).round(2)
     @value = 0.to_d if @value.sign == BigDecimal::SIGN_NEGATIVE_ZERO
     @cents = (@value * 100).to_i
+    freeze
   end
 
   def -@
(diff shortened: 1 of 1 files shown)


lib/money/money.rb
```diff
@@ -8,6 +8,16 @@ class Money
...
   class << self
     alias_method :from_amount, :new
+
+    def new(value = 0, _currency = nil)
+      if value == 0
+        @empty ||= from_amount(0).freeze
+      else
+        from_amount(value, _currency)
+      end
+    end
```
wvanbergen: I find it kind of weird that the 0 value is immutable, but other values aren't. When are we actually mutating these?

lib/money/money.rb
```diff
@@ -8,6 +8,16 @@ class Money
 
   class << self
     alias_method :from_amount, :new
+
+    def new(value = 0, _currency = nil)
+      if value == 0
+        @empty ||= from_amount(0).freeze
```
DazWorrall: Would it matter if if a currency was specified?

lib/money/money.rb
casperisfine: They are all immutable. It's just a matter of what range of values is worth caching. [Just like how the JVM caches Integer instances between -128 and 127](http://stackoverflow.com/a/3637974).

In our case, 0 is by a very far the most common value, other aren't really worth caching unless we want to do some kind of LRU, but it's taking it a bit far.

wvanbergen: In that case, should we call `freeze` in `from_amount` ?

lib/money/money.rb
casperisfine: `_curency` is currently a stub to provide compatibility with https://github.com/RubyMoney/money because we want to migrate to it.

So it doesn't really matter right now. But yes if we want to keep that perf patch once the transition is done, then we'll have to rethink it a bit, even though `0` in any currency is generally equal to 0 in any other currency.

lib/money/money.rb
casperisfine: sure.

casperisfine: done.

PR #51, opened 2017-05-02
Simple optimization of Money#allocate

casperisfine:
So I noticed `amounts_from_splits` quite high on the profiles:

capture d ecran 2017-05-02 a 17 36 39

Right now when you call allocate with Money instances as argument (rather common), the money instances are converted as follow: BigDecimal#to_f#to_s#to_d

It's kind of wasteful...

I changed the codepath so we return the `value` attribute directly since it's a BigDecimal already.

@wvanbergen @DazWorrall for review please.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -274,15 +274,21 @@ class Money
     splits.all? { |split| split.is_a?(Rational) }
   end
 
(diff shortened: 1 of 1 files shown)


DazWorrall:
Jus out of curiosity how are you profiling @casperisfine, or is this data from Newrelic?

PR #56, opened 2017-05-08
Revert "Revert "adds #to_d support to Money object""

elfassy:
Reverts Shopify/money#47

adds support for `to_d` to money objects


bdewater:
Why is the merge button so green and shiny 🤣

PR #57, opened 2017-05-19
Add conversion to string in Money#parse if input is not a string

clandry94:
**Why**

Inputs to `Money#parse` throw an error if they are anything besides a string. It would make sense for something like `Money.parse(1)` to work.

**What**

Added a simple cast to string if the input on parse is anything besides a string.

### DIFF
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -7,7 +7,7 @@ class MoneyParser
   end
   
   def parse(input)
-    Money.new(extract_money(input))
+    Money.new(extract_money(input.to_s))
   end
   
   private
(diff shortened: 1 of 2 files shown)


lib/money/money_parser.rb
```diff
@@ -7,6 +7,7 @@ def self.parse(input)
   end
   
   def parse(input)
+    input = input.to_s if !input.is_a?(String)
     Money.new(extract_money(input))
```
celsodantas: I think it makes sense to do a `Money.new(extract_money(input.to_s))` here and remove the line above.

celsodantas:
@Shopify/byroot mind to CR this too?

lib/money/money_parser.rb
clandry94: Updated!

PR #59, opened 2017-06-14
Currency support

elfassy:
# What
Add currency support to Money and output deprecation warnings when doing computation across currencies. Prepare for https://github.com/Shopify/money/issues/68

# Why
Having a currency aware Money object allows us to correctly represent minor units instead of simply multiplying or dividing by 100.


lib/money/currency.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+    currencies_json
+
+    def initialize(currency_iso)
+      return nil if currency_iso.blank?
+      unless data = self.class.currencies_json[currency_iso]
+        raise UnknownCurrency, "Unknown currency '#{currency_iso}'"
+      end
+      @alternate_symbols     = data['alternate_symbols']
+      @decimal_mark          = data['decimal_mark']
```
bdewater: As discussed today we should remove this and thousands_separator as these are properties from the locale and not from the currency.

lib/money_column/active_record_hooks.rb
```diff
@@ -5,11 +5,13 @@ def self.included(base)
     end
 
     module ClassMethods
```
bdewater: Maybe we should deprecate this entire class. AR provides built-in support for BigDecimal data types and with [composed_of](http://api.rubyonrails.org/v5.0.0/classes/ActiveRecord/Aggregations/ClassMethods.html) it can do the same thing money_column does, and more (using it in finder methods).

lib/money/core_extensions.rb
```diff
@@ -12,7 +12,11 @@ def to_money
 #   '100'.to_money => #<Money @cents=10000>
 #   '100.37'.to_money => #<Money @cents=10037>
 class String
-  def to_money
-    empty? ? Money.empty : Money.parse(self)
+  def to_money(currency = Money.default_currency)
+    empty? ? Money.empty : Money.parse(self, currency)
+  end
+
+  def to_currency
```
bdewater: I'm not too keen on monkey patching core classes and IMO `1.to_money('EUR')` is becoming dubious if it's meant as a convenience method. For `"EUR".to_currency` I don't see the added value.

lib/money/currency.rb
```diff
@@ -0,0 +1,72 @@
...
+    @@mutex = Mutex.new
+
+    attr_reader :priority, :iso_code, :iso_numeric, :name, :symbol,
+      :disambiguate_symbol, :html_entity, :subunit, :subunit_to_unit, :decimal_mark,
+      :thousands_separator, :symbol_first, :smallest_denomination, :alternate_symbols
+
+    class << self
+      def new(currency_iso)
+        iso = currency_iso.to_s.downcase
+        loaded_currencies[iso] || @@mutex.synchronize { loaded_currencies[iso] = super(iso) }
```
bdewater: Awesome you thought of multi-threaded loading 👍

lib/money/money.rb
```diff
@@ -1,32 +1,51 @@
...
       new(0)
     end
+    alias_method :zero, :empty
 
-    alias_method :from_amount, :new
+    def parse(input, _currency = default_currency)
+      parser.parse(input)
+    end
+
+    def from_cents(cents, currency = default_currency)
```
bdewater: Do we want to deprecate this and promote something like `from_subunits` instead? That one would be aware of currency instead of dumb division by 100.

lib/money/money.rb
```diff
@@ -1,32 +1,51 @@
...
-        super(value, _currency)
+    attr_accessor :parser, :default_currency
+
+    def new(value = nil, currency = default_currency)
+      if value.nil?
+        value = 0
+        deprecate("Support for Money.new(nil) will be removed from the next major revision. Please use Money.new(0) or Money.zero instead.\n")
       end
+      return super(value, currency) unless value == 0
+      @@zero_money || @@mutex.synchronize { @@zero_money = super(0) }
```
bdewater: Wouldn't this make more sense as a constant like DECIMAL_ZERO?

lib/money/core_extensions.rb
elfassy:
```
1.to_money('EUR') #=> Money.new(1, 'EUR')
vs
'EUR'.to_currency #=> Money::Currency.new('EUR')
```
I can remove the `to_currency` in favour of the using the class 👍

lib/money/money.rb
elfassy: Unfortunately we can't or we'll end up calling the constant while we are initializing it `@@zero_money -> new(0) -> if 0 -> @@zero_money`. We need access to super `new`

lib/money/money.rb
elfassy: 👍 👍

lib/money_column/active_record_hooks.rb
elfassy: That would require every table missing a currency column to add one. Let's add this deprecation notice in a future PR as to not crowd the logs too much

bdewater: If they don't have one it's kind of scary. By the looks of the docs they can use procs for `constructor` to use a default currency for those columns. But we can follow up on this later.

lib/money/money.rb
```diff
@@ -84,7 +84,9 @@ def -(other)
   end
 
   def *(numeric)
-    raise TypeError, "#{numeric} is not Numeric" unless numeric.is_a?(Numeric)
+    unless numeric.is_a?(Numeric)
+      Money.deprecate("Multiplying Money with #{numeric.class.name} is deprecated and will be removed in the next major release.")
```
bdewater: Before it was impossible to multiply with a Numeric, now it will be possible + a warning? This seems to achieve the opposite.

elfassy: `unless numeric.is_a?(Numeric)` So if it's not a numeric we show the deprecation warning

elfassy: you're looking at the diff between two commits of this PR. What we have now is the old behaviors (i had added the raise in this PR)

config/currency_iso_extra.json
```diff
@@ -0,0 +1,18 @@
+{
+  "xxx": {
```
Krystosterone: rather than separate file, can we have a test to ensure this is always there?

lib/money/currency.rb
```diff
@@ -0,0 +1,71 @@
...
+      @iso_code              = data['iso_code']
+      @iso_numeric           = data['iso_numeric']
+      @name                  = data['name']
+      @smallest_denomination = data['smallest_denomination']
+      @subunit_to_unit       = data['subunit_to_unit']
+      @minor_units           = subunit_to_unit == 0 ? 0 : Math.log(subunit_to_unit, 10).round.to_i
+      freeze
+    end
+
+    def ==(other)
```
Krystosterone: thinks of using `Forwardable`

lib/money/currency.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      def loaded_currencies
+        @@loaded_currencies ||= {}
+      end
+
+      def currencies_json
+        @@currencies_json ||= begin
+          currencies = {}
+          Dir.glob("#{CURRENCY_DATA_PATH}/*.json") do |currency_file|
```
Krystosterone: Single file loaded only when XXX is put into the same file

lib/money/currency.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    class << self
+      def new(currency_iso)
+        iso = currency_iso.to_s.downcase
+        loaded_currencies[iso] || @@mutex.synchronize { loaded_currencies[iso] = super(iso) }
+      end
+      alias_method :find!, :new
+
+      def find(currency_iso)
+        return nil if currency_iso.blank?
```
Krystosterone: Remove this and have `raise UnknownCurrency, "Invalid iso4217 currency '#{currency_iso}'"` in `def new`

spec/currency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it "is constructable with a symbol" do
+      expect(Money::Currency.new(:usd).iso_code).to eq('USD')
+    end
+
+    it "is constructable with a lowercase string" do
+      expect(Money::Currency.new('usd').iso_code).to eq('USD')
+    end
+
+    it "has currency data accessible" do
```
Krystosterone: This should be:
```
CURRENCY_DATA.keys.each do |attribute|
  describe "##{attribute}"
  ...
```

spec/currency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:currency_data) {{
+    "iso_code": "USD",
+    "name": "United States Dollar",
+    "subunit_to_unit": 100,
+    "iso_numeric": "840",
+    "smallest_denomination": 1,
+    "minor_units": 2
+  }}
+
+  describe "#new" do
```
Krystosterone: `".new"`

spec/currency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(Money::Currency.new('usd').iso_code).to eq('USD')
+    end
+
+    it "has currency data accessible" do
+      currency_data.keys.each do |attribute|
+        expect(currency.public_send(attribute)).to eq(currency_data[attribute])
+      end
+    end
+
+    it "raises if the currency is invalid" do
```
Krystosterone: Missing valid currency

spec/currency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect { Money::Currency.new('yyy') }.to raise_error(Money::Currency::UnknownCurrency)
+    end
+
+    it "raises if the currency is nil" do
+      expect { Money::Currency.new(nil) }.to raise_error(Money::Currency::UnknownCurrency)
+    end
+  end
+
+  describe "#find" do
```
Krystosterone: missing invalid currency

spec/currency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    it "raises if the currency is invalid" do
+      expect { Money::Currency.find!('yyy') }.to raise_error(Money::Currency::UnknownCurrency)
+    end
+
+    it "returns a valid currency" do
+      expect(Money::Currency.find!('CAD')).to eq(Money::Currency.new('CAD'))
+    end
+  end
+
+  describe "#eql?" do
```
Krystosterone: How about `==`

spec/money_spec.rb
```diff
@@ -54,6 +86,11 @@
     expect((Money.new + Money.new)).to eq(Money.new)
   end
 
+  it "adds a deprecation warning for inconsistent currencies" do
+    expect(Money).to receive(:deprecate).once
+    expect((Money.new(5, 'USD') + Money.new(1, 'CAD'))).to eq(Money.new(6, 'USD'))
```
Krystosterone: No need for `except`

lib/money/currency.rb
elfassy: I'm not really delegating to a different object so i'll use `alias_method` instead

spec/currency_spec.rb
elfassy: for new valid is split in 3 tests above:
`it "is constructable with a uppercase string" do` , `it "is constructable with a symbol" do`, `it "is constructable with a lowercase string" do`

spec/currency_spec.rb
elfassy: right below `it "returns nil if the currency is invalid" do`

elfassy:
@bdewater @Krystosterone this pr is ready for a final look, thx!

Krystosterone:
Following planning, should I hold off on 👀 this?

Krystosterone:
Isn't there a missing test to check for the `xxx` presence in the json file?

elfassy:
@Krystosterone it was tested but no explicitly. Added tests. Let me know if there's anything else. Thanks!

lib/money/money.rb
```diff
@@ -1,77 +1,107 @@
...
+
+    def from_subunits(subunits, currency_iso)
+      currency = Currency.find!(currency_iso)
+      value = Helpers.value_to_decimal(subunits) / currency.subunit_to_unit
+      new(value, currency)
+    end
+
+    def default_settings
+      self.parser = MoneyParser
+      self.default_currency = 'xxx'
```
sunblaze: `'xxx'` looks like a magic constant. Is this used to represent an unknown currency?

lib/money/money.rb
```diff
@@ -1,77 +1,107 @@
...
-
 class Money
   include Comparable
 
-  attr_reader :value, :cents
+  attr_reader :value, :cents, :currency
 
   class << self
-    def new(value = 0, _currency = nil)
+    attr_accessor :parser, :default_currency
```
sunblaze: I think the `default_currency` class attribute could be dangerous, if this gem is used by another web framework that supports threading, one thread could set the default currency and impact another separate thread. Do we need `default_curency` being settable?

sunblaze: I see the Money gem does this as well, but still wonder if this is necessary if we don't plan to use it immediately.

README.md
```diff
@@ -2,9 +2,97 @@
...
+By default `Money` defaults to Money::NullCurrency as its currency. This can be overwritten
+using:
+
+``` ruby
+Money.default_currency = Money::Currency.new("USD")
+```
+
+If you use Rails, then `environment.rb` is a very good place to put this.
+
+### Currency Exponent
```
sunblaze: Why mention exponent if we call the method `minor_units`?

lib/money/money.rb
```diff
@@ -1,77 +1,107 @@
...
+
+    def new(value = 0, currency = nil)
```
sunblaze: Why default to `nil` here, can you make the default `default_currency` in the method definition instead of the body? Or do we want to promote being able to pass a nil value in to indicate to use the default currency?

lib/money/core_extensions.rb
```diff
@@ -2,8 +2,8 @@
 #   100.to_money => #<Money @cents=10000>
 #   100.37.to_money => #<Money @cents=10037>
 class Numeric
-  def to_money
-    Money.new(self)
+  def to_money(currency = Money.default_currency)
```
sunblaze: This doesn't seem consistent with `Money.new`'s currency defaulting to `nil`. Either this should be match `Money.new` or make it match this format. Alternatively you could splat the args and pass any number of args to `Money.new`

lib/money/core_extensions.rb
```diff
@@ -12,7 +12,7 @@ def to_money
 #   '100'.to_money => #<Money @cents=10000>
 #   '100.37'.to_money => #<Money @cents=10037>
 class String
-  def to_money
-    empty? ? Money.empty : Money.parse(self)
+  def to_money(currency = Money.default_currency)
```
sunblaze: This doesn't seem consistent with `Money.new`'s currency defaulting to `nil`. Either this should be match `Money.new` or make it match this format. Alternatively you could splat the args and pass any number of args to `Money.new`

lib/money/currency.rb
```diff
@@ -0,0 +1,61 @@
...
+    @@loaded_currencies = {}
+
+    attr_reader :iso_code, :iso_numeric, :name, :smallest_denomination,
+                :subunit_to_unit, :minor_units, :symbol, :disambiguate_symbol
+
+    class << self
+      def new(currency_iso)
+        raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.empty?
+        iso = currency_iso.to_s.downcase
+        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
```
sunblaze: I'm a little confused what method is `super` in this context?

spec/null_currency_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+describe "NullCurrency" do
+  let (:null_currency) {Money::NullCurrency.instance}
+
+  it 'exposes the same public interface as Currency' do
+    expect(Money::NullCurrency).to quack_like Money::Currency
```
sunblaze: 🎉  nice interface testing!

lib/money/money.rb
elfassy: I'm defaulting to `nil` here because we also want to handle an explicit `nil` value coming from the DB's currency column. Airing on the side of caution. In other words, we want
```
Money.new(1, nil) == Money.new(1)
```

lib/money/money.rb
elfassy: added instructions in the readme on best practices. `Money.default_curency = Money::Currency.new('usd')` should be used in an initializer and not in the scope of a request

elfassy:
image

emilienoel:
Quick question on this: Why is nil currency == any currency?
screen shot 2018-05-11 at 1 25 01 pm

elfassy:
This was to allow for a slow rollout transition from money objects with no
concept of currency to ones with currency (since it was impossible to add a
currency to all money objects at once, and we didn't want to break existing
behaviour). We're getting close to a point we're we can raise if a money
object is created without a currency which will resolve this issue.

agarcher:
I recently got burned by [this](https://github.com/Shopify/money/pull/59#issuecomment-388430310). Are we at the point where we can drop support for `nil` currency?

PR #62, opened 2017-06-15
add missing tests

elfassy:
Get code coverage back to 100%


bdewater:
👍  can we get the suite to fail too if it dips below 100% again?

PR #65, opened 2017-06-19
deprecate the MoneyParser in favour or a really basic parser

elfassy:
# What
Deprecation warnings for the `MoneyParser` (to be removed in the next major release and moved to its own gem). https://github.com/Shopify/money/issues/70

# Why
A money string is dependent on the country and preferences and be as simple or as complex as people want. It is thus best to let people write their own parsers to decide on the rules they'd like implemented. Examples: `1.000` could be `one thousand` or `one` depending on the locale and/or currency.
This gem's parser should be much simpler than what it is right now and not make these assumptions.


elfassy:
related to https://github.com/Shopify/money/pull/43

PR #71, opened 2017-07-06
Add `Money.with_currency` similar to Rails' `Time.use_zone`

bdewater:
This allows to set the implicit default currency in a block scope, without having to set it explicitly inside the scope every time a new money object is created.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -23,8 +23,8 @@ class Money
     end
     alias_method :empty, :zero
 
-    def parse(input, _currency = nil)
-      parser.parse(input)
+    def parse(input, currency = nil)
+      parser.parse(input, currency)
     end
 
     def from_cents(cents, currency = nil)
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -7,7 +7,7 @@ class Money
     attr_accessor :parser, :default_currency
 
     def new(value = 0, currency = nil)
-      currency ||= default_currency
+      currency ||= current_currency || default_currency
 
       if value == 0
         @@zero_money ||= {}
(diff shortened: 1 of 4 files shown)


lib/money/money.rb
```diff
@@ -37,6 +37,19 @@ def from_subunits(subunits, currency_iso)
       new(value, currency)
     end
 
+    # Set Money.default_currency inside the supplied block, resets it to
+    # the previous value when done to prevent leaking state. Similar to
+    # ActiveSupport's Time.use_zone. This won't affect instances being created
+    # with explicitly set currency.
+    def use_currency(currency)
+      begin
+        old_currency, Money.default_currency = Money.default_currency, currency
```
elfassy: `Thread.current`

lib/money/money.rb
```diff
@@ -37,6 +37,19 @@ def from_subunits(subunits, currency_iso)
       new(value, currency)
     end
 
+    # Set Money.default_currency inside the supplied block, resets it to
+    # the previous value when done to prevent leaking state. Similar to
+    # ActiveSupport's Time.use_zone. This won't affect instances being created
+    # with explicitly set currency.
+    def use_currency(currency)
```
elfassy: what do you think of `with_currency` instead, I think it's closer to what's used elsewhere

spec/money_spec.rb
```diff
@@ -632,4 +632,43 @@
       Money.deprecate('ok')
     end
   end
+
+  describe('#use_currency') do
```
Krystosterone: Unnecessary `()`

spec/money_spec.rb
```diff
@@ -632,4 +632,43 @@
...
+    it "allows setting the implicit default currency for a block scope" do
+      money = nil
+      Money.with_currency('CAD') do
+        money = Money.new(1.00)
+      end
+
+      expect(money.currency.iso_code).to eq('CAD')
+    end
+
+    it "does not use the currency for a block scope when explicitly set" do
```
Krystosterone: 👏

lib/money/money.rb
```diff
@@ -37,6 +35,27 @@ def from_subunits(subunits, currency_iso)
       new(value, currency)
     end
 
+    def current_currency
+      Thread.current[:money_currency]
```
Krystosterone: 👏

PR #72, opened 2017-07-07
encode_with currency for simple serialization

elfassy:
# Why
There's no need to serialize complex objects such as the currency objects, only the iso_code string is needed. 
- use less resources (no need to store all the currency data on every record)
- easier for humans to read db records
- old gem versions will be able to deserialize the new format (otherwise might raise Currency class not found), useful for phased deploys

# What
- use `encode_with` to specify how to serialize the money object and `init_with` to deserialize
- serialize currency as the `iso_code` string

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -57,7 +57,7 @@ class Money
   end
 
   def encode_with(coder)
-    coder['value'] = @value
+    coder['value'] = @value.to_s('F')
     coder['currency'] = @currency.iso_code
   end
 
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -53,7 +53,12 @@ class Money
   end
 
   def init_with(coder)
-    initialize(coder.map['value'], coder.map['currency'])
+    initialize(coder['value'], coder['currency'])
+  end
+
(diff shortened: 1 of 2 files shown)


elfassy:
Added a second commit where we serialize value as float instead of keeping it as bigdecimal. I'm on the fence on this one would like to hear what you think. It does make the value human readable but it required converting to float and then back to bigdecimal when we deserialize.

sunblaze:
Isn't it a bad idea usually to convert money objects to floats? Can you serialize the BigDecimal to a string instead?

elfassy:
@sunblaze Initially didn't want to go to string because
```
BigDecimal.new(1.29, 15).to_s
=> "0.129E1"
```
but it turns out `to_s` takes an argument to show how we want it formatted https://ruby-doc.org/stdlib-1.9.3/libdoc/bigdecimal/rdoc/BigDecimal.html#method-i-to_s
```
BigDecimal.new(1.29, 15).to_s('F')
=> "1.29"
```
I'll update 👍

elfassy:
🔔  are we good with this PR?

PR #74, opened 2017-07-10
Handle deserialization of null currency

elfassy:
# What
- Handle deserialization of null currency
- keep currency if doing a computation with a null currency
- add missing helpers unit tests

# Why
We're now only saving the ISO code in the DB. This means that for NullCurrency we have `xxx`. When we deserialize this iso code gets passed to the currency class which needs to be able to handle this code properly. ~~We considered moving this logic to the `value_to_currency` helper, but decided on the Currency class in order to prevent someone doing a manual `Currency.find('xxx')` from raising an exception.~~

### DIFF
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -14,14 +14,7 @@ class Money
       def new(currency_iso)
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.empty?
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize do
-          @@loaded_currencies[iso] =
-            if iso == 'xxx'
-              NullCurrency.new
-            else
-              super(iso)
-            end
-        end
+        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
       end
       alias_method :find!, :new
 
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -14,7 +14,14 @@ class Money
       def new(currency_iso)
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.empty?
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        @@loaded_currencies[iso] || @@mutex.synchronize do
+          @@loaded_currencies[iso] =
+            if iso == 'xxx'
+              NullCurrency.new
+            else
+              super(iso)
+            end
+        end
       end
       alias_method :find!, :new
 
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -37,12 +37,16 @@ class Money
       when Money::Currency, Money::NullCurrency
         currency
       else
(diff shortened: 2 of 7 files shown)


lib/money/currency.rb
```diff
@@ -14,7 +14,9 @@ class << self
       def new(currency_iso)
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.empty?
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        @@loaded_currencies[iso] || @@mutex.synchronize do
+          @@loaded_currencies[iso] = iso == 'xxx' ? NullCurrency.new : super(iso)
```
sunblaze: Does NullCurrency need to be cached in the @@loaded_currencies?

elfassy: doesn't need to be, but why not do it?

sunblaze: Yeah, so I was also looking at how this one line is already pretty complicated. So if we can pull out the check for a NullCurrency to before the `@@mutex.synchronize` we save having to do a synchronize on the first NullCurrency and this line becomes easier to read.

```ruby
return NullCurrency.new if iso == 'xxx'
@@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
```

lib/money/currency.rb
```diff
@@ -14,7 +14,14 @@ class << self
...
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        @@loaded_currencies[iso] || @@mutex.synchronize do
+          @@loaded_currencies[iso] =
+            if iso == 'xxx'
+              NullCurrency.new
+            else
+              super(iso)
+            end
+        end
```
sunblaze: Yeah, so I was also looking at how this one line is already pretty complicated. So if we can pull out the check for a NullCurrency to before the `@@mutex.synchronize` we save having to do a synchronize on the first NullCurrency and this line becomes easier to read.

```ruby
return NullCurrency.new if iso == 'xxx'
@@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
```

elfassy: Easier to read but we'd lose memoization. Discussed with Krystian and agreed that it would be better to move this logic to the `value_to_currency` helper since a `.new` does not usually return nil (or in this case our null object).

elfassy:
@Krystosterone please give this a second look, thanks

lib/money/helpers.rb
```diff
@@ -37,12 +37,16 @@ def value_to_currency(currency)
       when Money::Currency, Money::NullCurrency
         currency
       else
-        if currency.nil? || currency.empty?
+        if currency.nil? || currency.empty? || null_currency?(currency)
```
Krystosterone: If it's a `null_currency?`, shouldn't it return a `Money::NullCurrency`?

Krystosterone: Also just realized, are there no unit tests for helpers?

elfassy: no, it's the same as `currency.nil?`, meaning that there's no currency setup

PR #66, opened 2017-06-19
add pry to gems for test debugging

elfassy:

PR #75, opened 2017-07-12
Improve test suite

bdewater:
This PR combines a bunch of things to make the test suite hard, better, faster (or at least as fast) and stronger:
- Adds a test to ensure old serialized currency YAML in Shopify is correctly handled for future people refactoring this code.
- Use a real currency that can cause deprecation warnings and defaulting to raise on deprecation. The first part is more reflective of real app usage and the second one is to make sure it's loud and clear if we make a mistake.
- RSpec config to disable monkey patching, verifying method existance and arity when stubbing, and only using expect syntax for future proofing.
- Random test ordering to catch mistakes we make in the test suite with before and after hooks.
- Add Pry and byebug to debug anything, supersedes https://github.com/Shopify/money/pull/66

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ pkg
 # For rubinius:
 #*.rbc
 Gemfile.lock
+
+.rspec_status
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -29,6 +29,11 @@ load File.join(File.dirname(__FILE__), "schema.rb")
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each {|f| require f}
 
 RSpec.configure do |config|
+  config.order = :random
+  
+  # Enable flags like --only-failures and --next-failure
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -47,3 +47,5 @@ pkg
 # For rubinius:
 #*.rbc
 Gemfile.lock
+
+.rspec_status
```
Krystosterone: linebreak plox

spec/spec_helper.rb
```diff
@@ -26,7 +31,20 @@
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each {|f| require f}
 
 RSpec.configure do |config|
+  config.order = :random
+  
+  # Enable flags like --only-failures and --next-failure
+  config.example_status_persistence_file_path = ".rspec_status"
+
+  config.disable_monkey_patching!
```
Krystosterone: Nice

PR #78, opened 2017-07-14
Add missing currencies

elfassy:
# Why
Some currencies are not valid iso currencies but are still used in "real life":tm:

# What
- currency_historic.json for all currencies that were once real currencies but are no longer
- currency_non_iso for currencies being used but are not part of the official iso standard (hint: they have a blank `iso_numeric` value) 
Example: 
> ISO 4217 does not include a separate currency code for the Jersey pound, but where a distinct code is desired JEP is generally used
- add deprecation warning instead of raise in the case of an invalid currency code.


lib/money/currency.rb
```diff
@@ -1,14 +1,13 @@
+require "money/currency/loader"
+
 class Money
   class Currency
-    class UnknownCurrency < ArgumentError; end
-
-    CURRENCY_DATA = "#{File.expand_path('../../../config', __FILE__)}/currency_iso.json".freeze
+    extend Loader
```
pi3r: Why extend the Loader? I mean, you could just do ` @@currencies ||= Loader.load_currencies`.

bdewater:
@jahfer we did, JEP was the only missing currency 😄 😭 😄

lib/money/currency/loader.rb
```diff
@@ -0,0 +1,22 @@
+class Money
```
pi3r: no unit tests at all for the loader?

Issue #68, opened 2017-07-04
Currency Support


elfassy:
- Support for passing a currency/amount
```
Money.new(10.00, 'USD')
```
- Allow a default currency
```
Money.default_currency = 'USD'
Money.new(10.00) == Money.new(10.00, 'USD')
```
- Raise on cross currency computation
```
Money.new(1, 'USD') + Money.new(1, 'CAD') #=> raise
```

- Deprecate `from_cents` in favor of `from_subunits`
Will divide by the value by the number of subunits, not always 100

- to_s will show the correct number of subunits
```
Money.new('100', 'JPY').to_s #=> '100'
Money.new('100', 'SLL').to_s #=> '100.000'

Money.new('100', 'JPY').to_s(:legacy_dollars) #=> '100.00'
Money.new('100', 'SLL').to_s(:legacy_dollars) #=> '100.00'
```

- subunits to format with currency
```
Money.new(100, 'JPY').subunits #=> 100
Money.new(100, 'USD').subunits #=> 10000
```

bdewater:
I think we can call this done 😄

PR #81, opened 2017-07-24
add support for subunits

elfassy:
# Why
We want deprecate `cents` in favor of currency aware `subunits`. This PR adds the ability to use either. Needed for https://github.com/Shopify/money/issues/68

# What
- Add `Money.new(1, 'USD').subunits` 
- rename `to_s(:minor_units)` to `to_s(:subunits)` for consistency

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -83,6 +83,10 @@ class Money
     coder['currency'] = @currency.iso_code
   end
 
+  def subunits
+    (value * currency.subunit_to_unit).to_i
+  end
+
   def -@
     Money.new(-value, currency)
   end
@@ -189,11 +193,7 @@ class Money
     case style
     when :legacy_dollars, nil
       sprintf("%.2f", value)
-    when :legacy_cents
-      (value * 100).round.to_i.to_s
-    when :minor_units
-      (value * currency.subunit_to_unit).round.to_i.to_s
-    when :major_units
+    when :amount
       sprintf("%.#{currency.minor_units}f", value)
     end
   end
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -190,10 +191,10 @@ def to_s(style = nil)
     when :legacy_dollars, nil
       sprintf("%.2f", value)
     when :legacy_cents
-      (value * 100).round.to_i.to_s
-    when :minor_units
-      (value * currency.subunit_to_unit).round.to_i.to_s
-    when :major_units
```
joshnuss: is `minor_units` & `major_units` no longer used?

lib/money/money.rb
```diff
@@ -190,10 +191,10 @@ def to_s(style = nil)
...
+      cents.to_s
+    when :subunits
```
joshnuss: At first I thought `subunits` was just like `cents`, but region agnostic.
But it's really more like `total_in_subunits` or `total_in_lowest_unit`.

lib/money/money.rb
elfassy: They were never used, it's something we added in the currency PR. Just renaming for consistency

lib/money/money.rb
elfassy: > it's really more like total_in_subunits or total_in_lowest_unit
Not sure what you mean here. subunits are just like cents for 2 digit currencies
```
Money.new(1, 'USD').cents == 100
Money.new(1, 'USD').subunits == 100
#But
Money.new(1, 'JPY').cents == 100
Money.new(1, 'JPY').subunits == 1
```

elfassy:
I'm starting to think there's no need for subunit styling options in `to_s`. What do you think of removing `to_s(:subunits)` and `to_s(:cents)` in favour of `.subunits.to_s` and `.cents.to_s`

PR #82, opened 2017-07-25
Remove `money_column` and document better alternatives

bdewater:
A previous attempt to make `money_column` use proper ActiveRecord APIs for performance reasons was https://github.com/Shopify/money/pull/49 but that had to be reverted because it broke a lot of tests in the main Shopify app. Instead we converted over to the ActiveRecord [attributes API](http://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html) on a per model basis.

However the recent currency work means we need to use [`composed_of`](http://api.rubyonrails.org/v5.1.0/classes/ActiveRecord/Aggregations/ClassMethods.html) instead to map the amount and currency to two database columns and vice versa, as the attributes API doesn't (yet?) work with multiple columns. This makes `money_column` a relic of the past, so it's time to remove it.


lib/money_column/active_record_hooks.rb
```diff
@@ -1,41 +0,0 @@
...
-          define_method("#{name}_before_type_cast") do
-            send(name) && sprintf("%.2f", send(name))
-          end
-
-          define_method("#{name}=") do |value|
-            if value.blank? || !value.respond_to?(:to_money)
-              write_attribute(name, nil)
-              nil
-            else
-              money = value.to_money
```
elfassy: my only worry is this line. the `to_money` on a string calls the money_parser. Removing this will remove this behavior

bdewater: I've updated the PR description with links to AR docs, the attributes API example has shows how the value can be normalized at the ORM layer. Or you do it at the controller layer (call `Money.parse` yourself or something like stronger_parameters). Either way there's no reason to keep this around.

README.md
```diff
@@ -106,6 +106,35 @@ Money::Currency.new("JPY").minor_units  # => 0
...
+
+While `money_accessor` can build you some accessor getters and setters, we 
+recommend to use the right APIs instead as they enable more expressive
+interactions and you get caching for free:
+
+```ruby
+class Order < ApplicationRecord
+  composed_of :total_price, 
+    class_name: 'Money', 
+    mapping: [%w(price value), %w(currency currency)]
```
sunblaze: The mapping doesn't look to line up with the table you defined above in your example, shouldn't this be `total_price` instead of `price`? Not sure if the naming is to avoid conflict with the attribute names too

README.md
```diff
@@ -106,6 +106,35 @@ Money::Currency.new("JPY").minor_units  # => 0
...
+end 
+```
+
+It can be used as such:
+```ruby
+Order.create(price: Money.new(3.50, 'USD'))
+Order.where(price: Money.new(9.99, 'CAD'))
```
sunblaze: Looks like some inconsistencies here too, is it `price` or `total_price`

PR #83, opened 2017-07-29
Update `money_column` to use ActiveRecords `composed_of`

bdewater:
Supersedes #82 - while the advice still stands to use `composed_of` for performance reasons, typing those out can get very tedious and `money_column` can help with that for the use cases we've seen in Shopify: multiple columns with the same currency and/or a column that is in a static currency.

I've experimented with adding default presence validations but it felt too prescriptive. Since it "just works" I've documented just that in both the readme and tests.

### DIFF
--- a/spec/money_column_spec.rb
+++ b/spec/money_column_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 class MoneyRecord < ActiveRecord::Base
   RATE = 1.17
   before_validation do
-    self.price_usd = Money.new(self[:price] * RATE, 'USD') if attribute_present?('price')
+    self.price_usd = Money.new(self[:price] * RATE, 'USD')
   end
 
   money_column :price
@@ -25,7 +25,7 @@ RSpec.describe 'MoneyColumn' do
   let(:subject) { MoneyRecord.new(price: money, prix: toonie) }
   let(:record) do
     subject.save
-    subject.class.find(subject.id)
+    subject.reload
   end
 
   it 'returns money with currency from the default column' do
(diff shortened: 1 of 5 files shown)


spec/money_column_spec.rb
```diff
@@ -1,75 +1,84 @@
...
-    expect(m.price).to eq(Money.new(1.01))
+RSpec.describe 'MoneyColumn' do
+  let(:amount) { 1.23 }
+  let(:currency) { 'EUR' }
+  let(:money) { Money.new(amount, currency) }
+  let(:toonie) { Money.new(2.00, 'CAD') }
+  let(:subject) { MoneyRecord.new(price: money, prix: toonie) }
+  let(:record) do
+    subject.save
+    subject.class.find(subject.id)
```
sunblaze: Wouldn't `subject.reload` be simpler?

spec/money_column_spec.rb
```diff
@@ -1,75 +1,84 @@
...
-    expect(m.price).to eq(nil)
+  it 'returns money with currency from the specified column' do
+    expect(record.prix).to eq(Money.new(2.00, 'CAD'))
   end
 
-  it "typecasts money with missing currency column" do
-    m = MoneyRecord.new(price: Money.new(1.01, 'cad'))
-    expect(m.price).to eq(Money.new(1.01, Money::NullCurrency.new))
+  it 'returns money with the hardcoded currency' do
+    expect(record.price_usd).to eq(Money.new(1.44, 'USD'))
```
sunblaze: Where is 1.44 coming from?

sunblaze: Ah, the before_validation callback, missed that, thank you

PR #84, opened 2017-08-01
Deprecate cents in favour of subunits

elfassy:
# Why
cents are not currency aware, which is especially important for currencies that have no minor units like the japanese yen.

re https://github.com/Shopify/shopify/issues/121143

# What
- deprecate `money.cents`
- modify allocation so it allocates subunits correctly
- added tests to verify specifically JPY

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -69,7 +69,7 @@ class Money
   def initialize(value = 0, currency = nil)
     raise ArgumentError if value.respond_to?(:nan?) && value.nan?
     @currency = Helpers.value_to_currency(currency)
-    @value = Helpers.value_to_decimal(value).round(2)
+    @value = Helpers.value_to_decimal(value).round(@currency.minor_units)
     @subunits = (@value * @currency.subunit_to_unit).to_i
     freeze
   end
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -70,7 +70,7 @@ def initialize(value = 0, currency = nil)
     raise ArgumentError if value.respond_to?(:nan?) && value.nan?
     @currency = Helpers.value_to_currency(currency)
     @value = Helpers.value_to_decimal(value).round(2)
```
sunblaze: Don't we also need to not always round the decimal value to 2 as well?

elfassy: was going to do that with #73 but yea this won't work with 3 digit currencies. Will cherry-pick that commit here

spec/money_spec.rb
```diff
@@ -341,6 +343,13 @@
     end
   end
 
+  describe "value" do
+    it "rounds to the number of minor units provided by the currency" do
+      expect(Money.new(1.1111, 'USD').value).to eq(1.11)
```
bdewater: Add a three-decimal currency for completeness?

elfassy:
@christianblais @SebastianSzturo This PR will now correctly round money objects. This might have an impact on `JPY` (for example: https://github.com/Shopify/shopify/issues/124792).

PR #86, opened 2017-08-01
Allow non-standard currencies to be normalized into known ones

bdewater:
The list is lifted from https://github.com/Shopify/active_utils/blob/master/lib/active_utils/currency_code.rb except for PES - according to https://www.currency-iso.org/dam/downloads/lists/list_three.xml it was withdrawn on 1986 so it's unlikely we'll see those in transactions. 

We should definitely look into upgrading our currency database because like #85 it turns out to be incomplete.

### DIFF
--- a/config/currency_normalization_map.json
+++ b/config/currency_normalization_map.json
@@ -1,0 +1,20 @@
+{
+  "ARN": "ARS",
+  "AZM": "AZN",
+  "CHP": "CLP",
+  "DHS": "AED",
+  "ECD": "XCD",
+  "JAD": "JMD",
+  "JYE": "JPY",
+  "KUD": "KWD",
+  "MZM": "MZN",
+  "NMP": "MXN",
+  "NTD": "TWD",
+  "RDD": "DOP",
+  "RM": "MYR",
+  "RMB": "CNY",
+  "SFR": "CHF",
+  "SID": "SGD",
+  "UKL": "GBP",
+  "WON": "KRW"
+}
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -4,13 +4,14 @@ class Money
   class Currency
     @@mutex = Mutex.new
     @@loaded_currencies = {}
+    @@currency_normalization_map = nil
 
     class UnknownCurrency < ArgumentError; end
 
     class << self
       def new(currency_iso)
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.empty?
-        iso = currency_iso.to_s.downcase
+        iso = normalize_currency(currency_iso)
         @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
       end
       alias_method :find!, :new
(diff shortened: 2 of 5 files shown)


lib/money/currency.rb
```diff
@@ -24,14 +25,28 @@ def find(currency_iso)
       def currencies
         @@currencies ||= Loader.load_currencies
       end
+
+      def normalize_currency(currency_code)
+        code = currency_code.to_s.downcase
+        return code if currencies.keys.include?(code)
+
+        code = currency_normalization_map[code.upcase]
```
sunblaze: Why not have the normalization map in a downcased format to avoid this `upcase` and `downcase` shuffle

bdewater: I want to revisit the entire currency database soon by having it pull data from currency-iso.org and generate JSON from there. It's incomplete and incorrect (e.g. HUF as a fractionless currency). I want to fix the casing there to be upcase because that's what is normally used.

sunblaze: 👍

lib/money/currency.rb
```diff
@@ -24,14 +25,28 @@ def find(currency_iso)
...
+      end
     end
 
     attr_reader :iso_code, :iso_numeric, :name, :smallest_denomination,
                 :subunit_to_unit, :minor_units, :symbol, :disambiguate_symbol
 
     def initialize(currency_iso)
       data = self.class.currencies[currency_iso]
-      raise UnknownCurrency, "Invalid iso4217 currency '#{currency_iso}'" unless data
+      raise UnknownCurrency, "Invalid currency '#{currency_iso}'" unless data
```
sunblaze: Does this ever get hit now with the check in `normalize_currency`

config/currency_normalization_map.json
```diff
@@ -0,0 +1,20 @@
...
+  "NMP": "MXN",
+  "NTD": "TWD",
+  "RDD": "DOP",
+  "RM": "MYR",
+  "RMB": "CNY",
+  "SFR": "CHF",
+  "SID": "SGD",
+  "UKL": "GBP",
+  "WON": "KRW"
+}
```
elfassy: Why not add these to the `currency_historic.json` file?

elfassy: Or the `currency_non_iso.json` file were never real currencies

bdewater: These are currencies where people refer to them often enough with another code while they mean a currency that has another valid code. A common example I've seen at Shopify is that "NTD" is entered while "TWD" is the correct one. By returning a TWD currency instance, the correct currency code will end up in the database/used for ActiveMerchant calls/etc. 

We should take a look at those files again and see if some of those currencies should be placed here. It's a bit of a grey area, e.g. JEP to GBP seems like the wrong thing because even though they're closely related they are different enough to warrant a separate entry.

bdewater:
After dealing with this stuff some more in the mean time, I've come to regard this kind of normalization as a bad thing and should be left up to the application if it's desired.

PR #87, opened 2017-08-03
support for currency_read_only money_column helper

elfassy:
# Why
Support for only reading the currency from the DB but not writing it. This helps with the transition period to currency. 

```ruby
money_column :price, currency_column: 'my_currency', currency_read_only: true
```

# What
- add the `currency_read_only` option
- Show a deprecation warning the money currency doesn't match what is being saved to the DB
- read creates a money object with the currency from the DB

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -5,29 +5,37 @@ module MoneyColumn
     end
 
     module ClassMethods
-      def money_column(*columns, currency_column: :currency, currency: false)
-        raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency_column && currency
+      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false)
+        raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency && currency_column
 
-        columns = columns.flatten
-        if currency_column
-          columns.each do |column|
+        if currency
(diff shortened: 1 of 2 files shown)


spec/money_column_spec.rb
```diff
@@ -8,7 +8,8 @@ class MoneyRecord < ActiveRecord::Base
 
   money_column :price
   money_column :prix, currency_column: :devise
-  money_column :price_usd, currency_column: false, currency: 'USD'
+  money_column :price_usd, currency: 'USD'
+  money_column :price4, currency_read_only: true
```
bdewater: Instead of adding another column, WDYT about using another AR class like the `MoneyWithValidation` specs?

lib/money_column/active_record_hooks.rb
```diff
@@ -5,29 +5,37 @@ def self.included(base)
     end
 
     module ClassMethods
-      def money_column(*columns, currency_column: :currency, currency: false)
-        raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency_column && currency
+      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false)
+        raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency && currency_column
```
sunblaze: We should have a test for this

PR #92, opened 2017-08-25
refactor money column

### LINKED ISSUE #91: Reload doesn't always work correctly with money_column
Just noticed this while reviewing https://github.com/Shopify/money/pull/90

We set an instance variable when the currency is fixed or is readonly, but we don't clear the instance variable on a `reload` of the model.

Here's an example test of the issue:
```ruby
original = MoneyRecord.create!(price_usd: Money.new(1, 'USD'))
copy = MoneyRecord.last
copy.update!(price_usd: Money.new(2, 'USD'))
original.reload
assert_equal Money.new(2, 'USD'), original.price_usd
```

@elfassy

elfassy:
# Why

Money column was missing some key features namely
- proper reload fixes #91
- legacy support for saving `1.23` strings and 1.23 floats in addition to money objects
- proper handling of invalid currencies from the DB
- changing the currency invalidates the money column cache
- make the currency column explicit (remove default name)

# What
Tests and moved the code in functions for better readability

### DIFF
--- a/spec/money_column_spec.rb
+++ b/spec/money_column_spec.rb
@@ -36,12 +36,39 @@ RSpec.describe 'MoneyColumn' do
     expect(record.price).to eq(Money.new(1.23, 'EUR'))
   end
 
+  it 'duplicating the record keeps the money values' do
+    expect(MoneyRecord.new(price: money).clone.price).to eq(money)
+    expect(MoneyRecord.new(price: money).dup.price).to eq(money)
+  end
+
   it 'returns money with currency from the specified column' do
     expect(record.prix).to eq(Money.new(2.00, 'CAD'))
   end
 
   it 'returns money with the hardcoded currency' do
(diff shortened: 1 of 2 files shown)


sunblaze:
This also looks to change the `currency_column` from defaulting to currency if not specified

elfassy:
@sunblaze that's correct, reducing the amount of magic and making things explicit

spec/money_column_spec.rb
```diff
@@ -44,6 +49,27 @@ class MoneyWithReadOnlyCurrency < ActiveRecord::Base
     expect(record.price_usd).to eq(Money.new(1.44, 'USD'))
   end
 
+  it 'returns money when the currency is invalid' do
```
sunblaze: Test name could be more descriptive, it doesn't mention that the money object's currency is the null currency when. Also shouldn't the test also assert that we can read the amount fine still?

elfassy:
@sunblaze please let me know if there's anything else, thx

PR #93, opened 2017-08-29
active record queries with money columns

elfassy:
# Why
We want to be able to do search using money objects for money_colums. 
```ruby
total = Money.new(1, 'USD')
Cart.where(total: total).to_sql
#=> "SELECT * FROM `carts` WHERE `total` = 1.0"
```
The decision was made to ignore the currency when doing a query because it is more likely someone will be searching for `Money.zero` or that db record with nil currency would be missing from search results.

# What
At this point `composed_of` is causing more problems than it solves. This PR removes it in favor of a custom implementation. The only thing really missing was the `accessor` to provide the active record hook.

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -30,7 +30,8 @@ module MoneyColumn
         raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency && currency_column
 
         if currency
-          currency_object = Money::Currency.find!(currency).to_s
+          currency_iso = Money::Currency.find!(currency).to_s
+          currency_read_only = true
         elsif currency_column
           clear_cache_on_currency_change(currency_column)
         else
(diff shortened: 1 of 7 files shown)


spec/money_column_spec.rb
```diff
@@ -60,15 +60,21 @@ class MoneyWithReadOnlyCurrency < ActiveRecord::Base
   it 'handles legacy support for saving floats' do
     record.update(price: 3.21, prix: 3.21)
     expect(record.price.value).to eq(3.21)
+    expect(record.price.currency_iso).to eq(currency)
     expect(record.price_usd.value).to eq(3.76)
+    expect(record.price_usd.currency_iso).to eq('USD')
     expect(record.prix.value).to eq(3.21)
+    expect(record.prix.currency_iso).to eq('CAD')
```
bdewater: We might want to remove the `prix` column now that we've made the column name explicit all the time, it's not different anymore.

lib/money_column/active_record_hooks.rb
```diff
@@ -28,58 +28,78 @@ def init_internals
...
           if money.blank?
             write_attribute(column, nil)
             return @money_column_cache[column] = nil
           end
-          money = money.to_money
-          currency_db = Money::Helpers.value_to_currency(currency_object || try(currency_column))
 
-          unless currency_db && currency_db.compatible?(money.currency)
-            Money.deprecate("[money_column] currency mismatch between #{currency_db} and #{money.currency}.")
+          currency_db = currency_iso || try(currency_column)
```
bdewater: Maybe call this `currency_source` because a statically defined currency code doesn't come from the DB.

spec/money_column_spec.rb
elfassy: it does have a different currency, i'll have a look 👍

elfassy:
after much back and forth, it's best to not implement this functionality for the time being as it may confuse developers as to the expected behaviour (will it also lookup the currency or not?). For now queries will require the dev to split and query the value and currency explicitly. We may want to reopen this in the future. 
An alternative approach suggested by Sean would be to an explicit `where_money` method.

PR #95, opened 2017-08-29
remove composed_of

elfassy:
# Why
Remove duplicate code paths. Also there's:

> It's just a shitty API that doesn't do half of what it needs to and has a ton of gotchas
-- Sean Griffin

It also leads to issues like
```
     TypeError:
       can't cast Money::Currency
```

# What
✂️


elfassy:
closing in favor of https://github.com/Shopify/money/pull/98

PR #96, opened 2017-08-30
Delegate more methods to @value BigDecimal

bdewater:
And use Forwardable. These are convenient when working in apps :)

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -1,7 +1,9 @@
 class Money
   include Comparable
+  extend Forwardable
 
   attr_reader :value, :subunits, :currency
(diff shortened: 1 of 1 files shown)


lib/money/money.rb
```diff
@@ -1,7 +1,9 @@
 class Money
   include Comparable
+  extend Forwardable
 
   attr_reader :value, :subunits, :currency
+  def_delegators :@value, :zero?, :nonzero?, :positive?, :negative?, :to_i, :to_f, :hash
```
elfassy: may want to delegate to `to_d` instead or @value, always feels weird to delegate to an instance varibable

bdewater: Not that weird, it's in [the docs](http://ruby-doc.org/stdlib-2.3.0/libdoc/forwardable/rdoc/Forwardable.html) 😄

PR #97, opened 2017-08-30
Add SimpleMoneyParser and to_s with_currency

bdewater:
Useful for APIs if all you got is a string. cc @xuorig


lib/money/money.rb
```diff
@@ -194,6 +194,8 @@ def to_s(style = nil)
     case style
     when :legacy_dollars, nil
       sprintf("%.2f", value)
+    when :with_currency
+      "#{value.to_s('F')} #{currency.to_s}"
```
elfassy: I think the default should have currency before value `USD 10.00`. Did you put it after to `to_f` and `to_i` still work correctly?

bdewater: Haha, that's actually a nice unexpected side effect.

I've looked around and there's no standardized notation for monetary values like ISO8601 for timestamps with timezone. Lacking that, I'm used to writing stuff down in [SI notation](https://en.wikipedia.org/wiki/International_System_of_Units#General_rules), e.g. 1 MHz. It also neatly lines up with the argument order when calling `Money.new`.

Or you could also say this notation betrays my European origins where this is customary ;)

lib/money/accounting_money_parser.rb
```diff
@@ -1,7 +1,11 @@
-class AccountingMoneyParser < MoneyParser
-  private
-  def extract_money(input)
-    # set () to mean negativity. ignore $
-    super(input.gsub(/\(\$?(.*?)\)/, '-\1'))
+class AccountingMoneyParser < DeprecatedMoneyParser
+  class << self
```
filipebarcos: it's a personal taste, so is `class << self` some kinda pattern used inside `money` gem?

I personally think that when we have 1-2 (few) class methods, I'd rather have `def self.my_method` than using `class << self`. I think this way you have a better visualization.

Again this is a personal taste 👅

lib/money/money_parser.rb
```diff
@@ -1,31 +0,0 @@
-class MoneyParser
```
elfassy: some people are directly using this class `MoneyParser.parse(...)`. I don't think we should remove it

lib/money/accounting_money_parser.rb
elfassy: @filipebarcos I don't think you can do true private class methods without `class << self`

filipebarcos: RIGHT! sorry, missed the `private`

lib/money/money_parser.rb
bdewater: If you're not doing a find/replace beforehand, Ruby will yell about a missing class. The interface stays the same so it's pretty harmless if you ask me?

elfassy: Instead of renaming the class to `DeprecatedMoneyParser` shouldn't we just add a deprecation warning when someone tries to use it? 
I'd be in favor of not adding that deprecation warning until we know what we'd like to use as an alternative.

bdewater: Fair enough, we can rename it later and leave it for now since this is not a full replacement.

lib/money/money.rb
filipebarcos: Indeed, there's no actual pattern for that. Should we have different outputs depending on which currency?

filipebarcos: the reason why I'm asking that is because I guess north americanas sorta expect that currency would come first.  Maybe we have a `format` method that I'm not aware of that users can pick the format they want...

lib/money/simple_money_parser.rb
```diff
@@ -0,0 +1,23 @@
+# Parses amount with optionally a space and currency code. Raises an error if
```
filipebarcos: I guess would be nice to have a better visualization what "optionally a space and currency code" means.
something like... all of this you wrote + actual example
```ruby
# "USD 10.00"
```

lib/money/accounting_money_parser.rb
```diff
@@ -1,7 +1,11 @@
...
-  private
-  def extract_money(input)
+  class << self
+
+    private
+
+    def extract_money(input)
     # set () to mean negativity. ignore $
     super(input.gsub(/\(\$?(.*?)\)/, '-\1'))
+    end
```
elfassy: i think you're missing some indentation here

spec/money_spec.rb
```diff
@@ -43,6 +43,12 @@
     expect(non_fractional_money.to_s(:legacy_dollars)).to eq("1.00")
   end
 
+  it "to_s with_currency style" do
+    expect(amount_money.to_s(:with_currency)).to eq("1.23 USD")
+    expect(non_fractional_money.to_s(:with_currency)).to eq("1.0 JPY")
```
filipebarcos: I guess, since this line is for a non-factional money, would be nicer to have `"1 JPY"` as an output for `non_factional_money.to_s(:with_currency)`

spec/simple_money_parser_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "parses amount with 3 decimals and 0 dinar amount" do
+      expect(described_class.parse("0.123 JOD")).to eq(Money.new(0.123, 'JOD'))
+    end
+
+    it "parses negative amount with 3 decimals and 0 dinar amount" do
+      expect(described_class.parse("-0.123 JOD")).to eq(Money.new(-0.123, 'JOD'))
+    end
+
+    it "rounds too many decimals" do
+      expect(described_class.parse("1.235 EUR")).to eq(Money.new(1.24, 'EUR'))
```
filipebarcos: it's always a round up? Should we include examples like 
```ruby
expect(described_class.parse("1.231 EUR")).to eq(Money.new(1.23, 'EUR')) 
expect(described_class.parse("1.238 EUR")).to eq(Money.new(1.24, 'EUR'))
```

lib/money/money.rb
```diff
@@ -191,6 +191,8 @@ def to_s(style = nil)
     case style
     when :legacy_dollars
       sprintf("%.2f", value)
+    when :with_currency
+      "#{value.to_s('F')} #{currency.to_s}"
```
elfassy: Is there a use case for this now? It's fairly easy for someone to do this on their own and most likely this needs to be locale specific. In any case, if there is a use case you'd need
```
"#{to_s(:amount)} #{currency.to_s}"
```
to have the correct number of decimals

spec/simple_money_parser_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+
+RSpec.describe SimpleMoneyParser do
+  describe "amounts without currency" do
+    it "raises an error for an empty string" do
+      expect { described_class.parse("") }.to raise_error(SimpleMoneyParser::ParseError)
+    end
+
+    it "raises an error for an invalid string" do
+      expect { described_class.parse("no money") }.to raise_error(SimpleMoneyParser::ParseError)
```
elfassy: please add a test where this raises with a currency being passed

lib/money/simple_money_parser.rb
```diff
@@ -0,0 +1,23 @@
...
+  REGEX = /^(-?\d*\.?\d*)(?: ([A-Z]{3}))?/
+
+  class << self
+    def parse(input)
+      input = input.to_s
+      amount, currency = input.scan(REGEX).first
+
+      if amount.empty?
+        raise ParseError, "empty amount: #{input[0,32]}"
+      else
```
elfassy: no need for the `else`, just `end` will do

spec/simple_money_parser_spec.rb
bdewater: Yes, it's the default BigDecimal ROUND_HALF_UP behaviour

PR #98, opened 2017-08-30
Money column remove composed of

elfassy:
# Why
Remove `composed_of` in favor of a custom implementation. This makes sure we're always seeing the deprecation warnings correctly. It will also prevent us from updating the currency column with an `''` when saving a float or string instead of a money object.

Subset of https://github.com/Shopify/money/pull/93

# What
- Reduce the amount of logs by only having deprecation warnings when writing a money object. 
- Saving a string or float won't update the currency column
- improved tests

### DIFF
--- a/spec/money_column_spec.rb
+++ b/spec/money_column_spec.rb
@@ -122,6 +122,19 @@ RSpec.describe 'MoneyColumn' do
     end
   end
 
+  describe 'missing money_column currency arguments' do
+    let(:subject) do
+      class MoneyWithWrongCurrencyArguments < ActiveRecord::Base
+        self.table_name = 'money_records'
+        money_column :price
+      end
+    end
+
+    it 'raises an ArgumentError' do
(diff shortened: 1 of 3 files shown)


spec/money_column_spec.rb
```diff
@@ -36,6 +36,13 @@ class MoneyWithReadOnlyCurrency < ActiveRecord::Base
     expect(record.price).to eq(Money.new(1.23, 'EUR'))
   end
 
+  it 'writes the currency to the db' do
+    expect(Money).to receive(:deprecate).once
+    record.update(price: Money.new(4, 'JPY'))
+    expect(record.price.value).to eq(4)
+    expect(record.price.currency.to_s).to eq('JPY')
```
sunblaze: This test doesn't look to confirm the currency is written to the DB, can you load a fresh instance of the object or something after saving?

PR #99, opened 2017-09-05
money_column coerce_null option

elfassy:
# Why
In some cases nils are undesired and are expected to be 0$. A good use case for this would be taxes. 

# What
- Reader methods return Money.zero if coerce_null is true and the value is nil. 
- Default to `coerce_null: false`

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -26,7 +26,7 @@ module MoneyColumn
     end
 
     module ClassMethods
-      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false)
+      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false, coerce_null: false)
         raise ArgumentError, 'cannot set both currency_column and a fixed currency' if currency && currency_column
 
         if currency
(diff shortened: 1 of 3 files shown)


lib/money_column/active_record_hooks.rb
```diff
@@ -26,7 +26,7 @@ def init_internals
     end
 
     module ClassMethods
-      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false)
+      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false, allow_nil: true)
```
bdewater: Naming wise, what do you think about `coerce_null`? It would also require changing the code in this PR but I think it's a bit more clear on what will happen.

spec/money_column_spec.rb
```diff
@@ -21,6 +21,12 @@ class MoneyWithReadOnlyCurrency < ActiveRecord::Base
   money_column :price, currency_column: 'currency', currency_read_only: true
 end
 
+class MoneyRecordAllowNilFalse < ActiveRecord::Base
```
sunblaze: Can you rename this class AllowNilFalse doesn't make any sense now

PR #101, opened 2017-09-12
Money column with a delegated currency

elfassy:
# Why
Some people may want to use a delegated currency column. In those cases it's possible to get into a race condition between the money object and the associated record. For example
```ruby
class MoneyRecord < ActiveRecord::Base
  belongs_to :delegated_record
  delegate :currency, to: :delegated_record
  money_column :price, currency_column: 'currency', currency_read_only: true
end

record = MoneyRecord.new(price: Money.new(1), delegated_record: DelegatedRecord.new(currency: 'USD'))
#=> Module::DelegationError
# MoneyRecord#currency delegated to delegated_record.currency, but associated_record is nil

# VS

record = MoneyRecord.new(delegated_record: DelegatedRecord.new(currency: 'USD'), price: Money.new(1))
record.price #=> Money.new(1, 'USD')

```
This PR fixes this bug.

NOTE: all this is only needed because we want to show a deprecation warning when currencies are incompatible. This must be done in the writer method (since the reader method will fetch the correct currency). This means that when the currency is not yet available (next on the list of params), we'll not get a deprecation warning even though we should have. Can't think of a way around that since we can't reorder the params being associated to the record on the fly. Not having a deprecation warning is better than raising in this case.

# What

- rescue from `Module::DelegationError` in writer method and assume nil currency
- reader method will read currency and assign it (and take care of memoization)

### DIFF
--- a/lib/money/null_currency.rb
+++ b/lib/money/null_currency.rb
@@ -28,7 +28,11 @@ class Money
     def to_s
       ''
     end
-    
+
+    def no_currency?
+      true
+    end
+
     alias_method :==, :eql?
   end
 end
--- a/lib/money/null_currency.rb
+++ b/lib/money/null_currency.rb
@@ -29,10 +29,6 @@ class Money
       ''
     end
 
-    def no_currency?
-      true
-    end
-
     alias_method :==, :eql?
   end
 end
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -58,37 +58,31 @@ module MoneyColumn
           return @money_column_cache[column] if @money_column_cache[column]
           value = read_attribute(column)
           return if value.nil? && !coerce_null
-          iso = currency_iso || try(currency_column)
+          iso = currency_iso || send(currency_column)
           @money_column_cache[column] = Money.new(value, iso)
         end
       end
 
       def money_column_writer(column, currency_column, currency_iso, currency_read_only)
         define_method "#{column}=" do |money|
+          @money_column_cache[column] = nil
+
           if money.blank?
             write_attribute(column, nil)
-            return @money_column_cache[column] = nil
+            return nil
+          elsif !money.is_a?(Money)
+            return write_attribute(column, money)
           end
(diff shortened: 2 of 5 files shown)


spec/money_column_spec.rb
```diff
@@ -91,6 +97,21 @@ class MoneyRecordCoerceNull < ActiveRecord::Base
     expect(record.prix.currency.to_s).to eq('CAD')
   end
 
+  it 'does not overwrite a currency column with a null currency' do
+    Money.with_currency(Money::NullCurrency.new) do
```
elfassy: I can remove this `with_currency` here once https://github.com/Shopify/money/pull/102 is merged

lib/money_column/active_record_hooks.rb
```diff
@@ -65,30 +65,31 @@ def money_column_reader(column, currency_column, currency_iso, coerce_null)
...
-            return @money_column_cache[column] = Money.new(money, currency_object)
+            return money
           end
 
-          if currency_source && !currency_object.compatible?(money.currency)
-            Money.deprecate("[money_column] currency mismatch between #{currency_object} and #{money.currency}.")
+          currency_raw_source = begin
+            currency_iso || try(currency_column)
+          rescue Module::DelegationError
+            nil
```
bdewater: Ugly but needed for the mismatch.

lib/money/null_currency.rb
```diff
@@ -28,7 +28,11 @@ def eql?(other)
     def to_s
       ''
     end
-    
+
+    def no_currency?
```
sunblaze: `no_currency?` sounds pretty vague and not clear what it's used for, can we name this method something else? Like don't overwrite, don't write or read only

elfassy: `@money.no_currency?` works to me. `@currency.no_currency?` doesn't work though. Will change to only keep the one in money. Does that work for you?

elfassy:
@sunblaze good idea, added some tests, let me know how this looks to you know, thanks!

PR #102, opened 2017-09-12
Money.zero is uses null currency instead of default currency

elfassy:
# Why
We want to avoid writing a default currency in the DB if Money.zero is saved to a money_column. For example, we want:
```ruby
Money.default_currency = 'CAD'
object.currency #=> 'USD'
object.price = Money.zero
object.currency #=> 'USD'
```

# What
Modify `Money.zero` to have the currency be `NullCurrency` irregardless of the default currency

### DIFF
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -3,17 +3,14 @@ require 'yaml'
 
 RSpec.describe "Money" do
 
-  before(:each) do
-    @money = Money.new
-  end
-
   let (:money) {Money.new(1)}
   let (:amount_money) { Money.new(1.23, 'USD') }
   let (:non_fractional_money) { Money.new(1, 'JPY') }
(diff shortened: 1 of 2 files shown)


spec/money_spec.rb
```diff
@@ -3,24 +3,29 @@
 
 RSpec.describe "Money" do
 
-  before(:each) do
-    @money = Money.new
-  end
-
+  let (:money) {Money.new(1)}
```
bdewater: 👍  for removing this minitest-ism but by changing the value the test description "defaults to 0 when constructed with no arguments" doesn't match anymore.

elfassy: good catch, missed that one, fix incoming

elfassy: done

PR #105, opened 2017-09-13
consistent zero

elfassy:
# Why
It is possible to have `Money.new(0) != Money.zero` because a no null default currency is used. This can lead to some confusion. We still want the ability to have Money.new(0, 'USD') which is used during computations that sum to 0.

# What
- Make `Money.new(0)` and `Money.zero` have the exact same behavior
- Allow explicit `Money.new(0, 'USD')`
- Allow zero sum computation to retain currency `Money.new(1, 'USD') - Money.new(1, 'USD') == Money.new(0, 'USD')`

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -9,19 +9,13 @@ class Money
     attr_accessor :parser, :default_currency
 
     def new(value = 0, currency = nil)
-      currency ||= current_currency || default_currency
-
-      if value == 0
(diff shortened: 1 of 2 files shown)


elfassy:
Discussed with Bart, and we came to the conclusion that most people would expect to have `Money.new(0)` have the default currency just like any other `.new`. `.zero` on the other hand, is a shorthand for having no currency that people opt in.

PR #108, opened 2017-09-14
active record type serialization

elfassy:
# Why
We want to be able to do, without having to get the decimal value first
```ruby
Record.where(price: Money.new(1, 'USD'))
```

# What
- it will automagically convert money object to decimal for searching on the DB
- it will **NOT** search for the currency

# Notes

This change is only compatible with Rails 5+

### DIFF
--- a/lib/money_column/active_record_type.rb
+++ b/lib/money_column/active_record_type.rb
@@ -1,0 +1,6 @@
+class MoneyColumn::ActiveRecordType < ActiveRecord::Type::Decimal
+  def serialize(money)
+    return nil unless money
+    super(money.to_d)
+  end
+end
--- a/lib/money_column.rb
+++ b/lib/money_column.rb
@@ -1,2 +1,3 @@
 require_relative 'money_column/active_record_hooks'
+require_relative 'money_column/active_record_type'
 require_relative 'money_column/railtie'
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -39,6 +39,7 @@ module MoneyColumn
         end
 
         columns.flatten.each do |column|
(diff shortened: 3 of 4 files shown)


lib/money_column/active_record_type.rb
```diff
@@ -0,0 +1,6 @@
+class MoneyColumn::ActiveRecordType < ActiveRecord::Type::Decimal
+  def serialize(money)
+    return nil unless money
+    money.to_d
```
bdewater: `super(money.to_d)` so we hook into whatever AR is providing for us (now or in the future)?

elfassy:
@jasonwebster would love to hear your thoughts on this, thx!

jasonwebster:
I think this is generally a great idea. 

Why'd you end up using [`define_attribute`](http://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-define_attribute) instead of [`attribute`](http://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute)? It would be great to include an explanation of why in the commit description.

elfassy:
Originally went with `define_attribute` to use the low level api. Now that I'm giving this a second thought it's probably wiser to use `attribute` to avoid weird cases where the schema needs to be reloaded.
Any thoughts on making this gem rails5+? We could potentially move the currency_column code to its own gem to keep money compatible with older versions of rails. Likely not needed though since rails4 is quickly becoming unsupported.

bdewater:
> Any thoughts on making this gem rails5+? 

👍 

> We could potentially move the currency_column code to its own gem

All our apps should be on this version anyway, and this gem is not published in any form yet. Let's not pre-emptively take on a maintenance burden.

elfassy:
ping I think we're good with this PR. @jasonwebster let me know if this looks good to you, thx

PR #111, opened 2017-09-22
build rational with two money objects

elfassy:
# Why
Can't use decimals (they give an approximation)
```ruby
     Failure/Error: expect(Rational(Money.new(10.0).value, Money.new(15.0).value)).to eq(Rational(2,3))

       expected: (2/3)
            got: 0.666666666666666667 (#<BigDecimal:7fbc4594b408,'0.6666666666 66666667E0',18(36)>)
```
AND
can't use `subunits` because they are currency aware
```ruby
     Failure/Error: expect(Rational(Money.new(10.0, 'USD').subunits, Money.new(15.0, 'JPY').subunits)).to eq(Rational(2,3))

       expected: (2/3)
            got: (200/3)
```
# What
- new `Money.rational(money1, money2)` method to correctly handle any currency

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -39,6 +39,13 @@ class Money
       new(value, currency)
     end
 
+    def rational(money1, money2)
+      money1.send(:arithmetic, money2) do
+        factor = money1.currency.subunit_to_unit * money2.currency.subunit_to_unit
+        Rational((money1.value * factor).to_i, (money2.value * factor).to_i)
+      end
+    end
+
     def current_currency
       Thread.current[:money_currency]
     end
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -39,6 +39,13 @@ def from_subunits(subunits, currency_iso)
       new(value, currency)
     end
 
+    def rational(money1, money2)
+      money1.send(:arithmetic, money2) do
```
bdewater: Just a question, why send instead of a normal call?

elfassy: `send` allows us to access that private method. The alternative would be to make that method public which we don't really want or to make rational an instance method `money1.rational(money2)` (i think it's cleaner as a class method since it resembles more the `Rational(1,2)` syntax)

PR #112, opened 2017-09-25
allocate between money with and without currency

elfassy:
# Why
We need to handle cases where we're `allocate_max_amounts` with a money objects that have no currency while the money being allocated does (and vice versa)

# What
- when maximum amounts are integers or have no currency, correctly use the same currency as the allocated amount
- when the amount being allocated doesn't have a currency, use the currency from the max amount (if all max amounts are in the same currency)
- when there is a missmatch of currencies, raise an argument error

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -276,29 +276,33 @@ class Money
   #   Money.new(100).allocate_max_amounts([Money.new(5), Money.new(2)])
   #     #=> [Money.new(5), Money.new(2)]
   def allocate_max_amounts(maximums)
-    maximums = maximums.map { |max_amount| max_amount.to_money }
-    maximums_total = maximums.sum
+    allocation_currency = extract_currency(maximums + [self])
+    maximums = maximums.map { |max| max.to_money(allocation_currency) }
+    maximums_total = maximums.reduce(Money.new(0, allocation_currency), :+)
 
     splits = maximums.map do |max_amount|
-      next(Money.empty) if maximums_total.zero?
-      max_amount.value / maximums_total.value
+      next(0) if maximums_total.zero?
+      Money.rational(max_amount, maximums_total)
     end
 
-    total_allocatable = [subunits, maximums_total.subunits].min
+    total_allocatable = [
+      value * allocation_currency.subunit_to_unit,
+      maximums_total.value * allocation_currency.subunit_to_unit
+    ].min
 
     subunits_amounts, left_over = amounts_from_splits(1, splits, total_allocatable)
 
     subunits_amounts.each_with_index do |amount, index|
       break unless left_over > 0
 
-      max_amount = maximums[index].subunits
+      max_amount = maximums[index].value * allocation_currency.subunit_to_unit
       next unless amount < max_amount
 
       left_over -= 1
       subunits_amounts[index] += 1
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -361,10 +368,14 @@ def arithmetic(money_or_numeric)
...
+    block_given? ? yield(other) : other
   end
 
   def calculated_currency(other)
     no_currency? ? other : currency
   end
+
+  def all_currencies(money_array)
+    (money_array + [self]).reject(&:no_currency?).map(&:currency).uniq
+  end
```
bdewater: This method extraction doesn't make much sense IMO.

elfassy: changed it around a little, let me know what you think, thx :)

lib/money/money.rb
```diff
@@ -269,29 +271,33 @@ def allocate(splits)
   #   Money.new(100).allocate_max_amounts([Money.new(5), Money.new(2)])
   #     #=> [Money.new(5), Money.new(2)]
   def allocate_max_amounts(maximums)
-    maximums = maximums.map { |max_amount| max_amount.to_money }
-    maximums_total = maximums.sum
+    allocate_currency = extract_currency(maximums + [self])
+    maximums = maximums.map { |max| max.to_money(allocate_currency) }
```
felix-d: Is it a problem if self is a `NullCurrency` and `maximums` are not?

lib/money/money.rb
```diff
@@ -269,29 +271,33 @@ def allocate(splits)
...
+    maximums_total = maximums.reduce(Money.new(0, allocate_currency), :+)
```
felix-d: ❤️

spec/money_spec.rb
```diff
@@ -493,6 +493,22 @@
...
     end
 
+    specify "#allocate_max_amounts legal computation with no currency objects" do
+      expect(
+        Money.new(3075, 'JPY').allocate_max_amounts([2600, 475]),
+      ).to eq([Money.new(2600, 'JPY'), Money.new(475, 'JPY')])
+
+      expect(
+        Money.new(3075, Money::NullCurrency.new).allocate_max_amounts([Money.new(2600, 'JPY'), Money.new(475, 'JPY')]),
+      ).to eq([Money.new(2600, 'JPY'), Money.new(475, 'JPY')])
```
felix-d: Nevermind seems it's not a problem, indeed both are converted to subunits using the same multiplier 👍

lib/money/money.rb
felix-d: If the amount is already a money object it won't do anything right?

elfassy: correct 👍 , this will only convert integer and strings to money objects.

lib/money/money.rb
```diff
@@ -21,7 +23,7 @@ def new(value = 0, currency = nil)
     alias_method :from_amount, :new
 
     def zero
-      new(0, NullCurrency.new)
+      new(0, NULL_CURRENCY)
```
sunblaze: This optimization is cool to see, but I don't think it relates to the issue with `allocate_max_amounts`. Would of been nice if this was in a separate PR or commit at least.

PR #114, opened 2017-09-26
Added fractional? and non_fractional? to Money::Currency

filipebarcos:
~~@bdewater this is was I was taking about in https://github.com/Shopify/shopify/pull/131505~~

This started off a conversation/suggestion in PR https://github.com/Shopify/shopify/pull/131505. The idea is to provide these two helper methods that will make refactoring easier across Shopify core code base, wherever we use the money gem.

### DIFF
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -58,5 +58,9 @@ class Money
     def fractional?
       @subunit_to_unit != 1
     end
+
+    def non_fractional?
+      !fractional?
+    end
   end
 end
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -54,5 +54,9 @@ class Money
 
     alias_method :==, :eql?
     alias_method :to_s, :iso_code
+
+    def fractional?
+      @subunit_to_unit != 1
+    end
   end
 end
--- a/lib/money/null_currency.rb
+++ b/lib/money/null_currency.rb
@@ -29,6 +29,10 @@ class Money
       ''
     end
 
+    def fractional?
+      true
+    end
+
     alias_method :==, :eql?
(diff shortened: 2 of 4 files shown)


bdewater:
👍  thanks for quickly jumping on this.

Unless my memory is mistaken (cc @elfassy) we're mostly needing this kind of check if the currency is non-fractional. What do you think about changing it to `nonfractional?` so it can be used like this:

```ruby
if currency.nonfractional?
  do_special_thing
else
  do_normal_thing
end

**Comment by filipebarcos** [comment]
@bdewater That was my initial reflex, but I was not sure about how we were mostly using.

Although, the idea behind using afirmative is that I find better when we use it for negative statements, like `if !non_fractional?` or `unless non_factional?`. To sum up, negating twice to make something truthy is not idea, imo.

If what you saying is true, then I'm +1 to changing it

**Comment by elfassy** [comment]
Not sure I see why would we want to call `money.non_fractional?`. The goal is to remove those methods.
This used to be needed because we didn't have proper knowledge of the currency within money. Now instead of 
```
currency.nonfractional? money.cents / 100 : money.cents
```
we have a better
```
money.subunits
```
which also handles other edge cases like `three_decimal_currency?(currency)`

**Comment by filipebarcos** [comment]
```
currency.nonfractional? money.cents / 100 : money.cents
```
is this the only kinda cases we need subunits checks? If that's the case, I agree that less checks (ifs) and do better abstractions 👍 

The goal here was to provide a better abstraction than doing manual subunit checking, but maybe there's a better abstraction

filipebarcos:
Closing this in favor of https://github.com/Shopify/shopify/pull/132169

PR #116, opened 2017-10-03
to_s with correct number of decimals based on currency

elfassy:
# Why
Follow up from https://github.com/Shopify/money/pull/115

Now that money is currency aware, we want
```ruby
Money.new(1, 'USD').to_s #=> '1.00'
Money.new(1, 'USD')      #=> #<Money value:1.00 currency:USD>
Money.new(1, 'JPY').to_s #=> '1'
Money.new(1, 'JPY')      #=> #<Money value:1 currency:JPY>
```

Re: https://github.com/Shopify/shopify/issues/121143

# What
Change the default string representation which also changes how `inspect` displays the result

# Note
It's still possible to get the old behavior by using `to_s(:legacy_dollars)`

### DIFF
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -34,8 +34,9 @@ RSpec.describe "Money" do
     end
   end
 
-  it "to_s as a float with 2 decimal places" do
+  it "to_s correctly displays the right number of decimal places" do
     expect(money.to_s).to eq("1.00")
+    expect(non_fractional_money.to_s).to eq("1")
   end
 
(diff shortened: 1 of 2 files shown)


spec/money_spec.rb
```diff
@@ -36,6 +36,7 @@
 
   it "to_s as a float with 2 decimal places" do
     expect(money.to_s).to eq("1.00")
+    expect(non_fractional_money.to_s).to eq("1")
```
bdewater: Test description doesn't match tests anymore.

PR #117, opened 2017-10-03
clamp can take params out of order

elfassy:
# Why

It's possible for code to mixup the min and max params to the clamp method. There's no reason to raise in this case since the clamp will work just as well.

```ruby
money.clamp(max, min)
# vs
money.clamp(min, max)
```

# What 
- remove raise if params are out not sorted
- improve tests

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -335,12 +335,10 @@ class Money
   # closest min or max value.
   #
   # @example
-  #   Money.new(50, "CAD").clamp(1, 100) #=> Money.new(10, "CAD")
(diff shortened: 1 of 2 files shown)


bdewater:
I'm +1 on fixing docs and tests but my original intent was to mirror [Ruby 2.4's clamp](https://ruby-doc.org/core-2.4.2/Comparable.html#method-i-clamp) where this doesn't work and returns the same error message.

elfassy:
Interesting that ruby's clamp doesn't allow inverting min and max. I guess it's a safety measure to verify intent. Closing for now, will re-open if this comes up again.

PR #119, opened 2017-10-05
Correction to Money#clamp rubydoc example

chrisbutcher:
Given the description `Clamps the value to be within the specified minimum and maximum. Returns self if the value is within bounds.` this example was incorrect.


bdewater:
Thanks!

PR #121, opened 2017-10-10
allow number strings with whitespace padding

elfassy:
# Why
`BigDecimal` correctly handles strings with whitespace padding therefore there's no reason to deprecate that as an invalid string.

```ruby
> BigDecimal.new(' 1.23 ').to_f
=> 1.23
```

# What
add any number of space `/s*` to the numeric validation regex

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -4,7 +4,7 @@ class Money
   module Helpers
     extend self
 
-    NUMERIC_REGEX = /\A-?\d*\.?\d*\z/.freeze
+    NUMERIC_REGEX = /\A\s*-?\d*\.?\d*\s*\z/
     DECIMAL_ZERO = BigDecimal.new(0).freeze
 
     def value_to_decimal(num)
--- a/spec/helpers_spec.rb
+++ b/spec/helpers_spec.rb
@@ -34,9 +34,21 @@ RSpec.describe Money::Helpers do
       expect(subject.value_to_decimal('1.23')).to eq(amount)
     end
 
+    it 'returns the bigdecimal version of a ruby number string with whitespace padding' do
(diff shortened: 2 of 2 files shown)


lib/money/helpers.rb
```diff
@@ -4,7 +4,7 @@ class Money
   module Helpers
     extend self
 
-    NUMERIC_REGEX = /\A-?\d*\.?\d*\z/.freeze
+    NUMERIC_REGEX = /\A\s*-?\d*\.?\d*\s*\z/
```
filipebarcos: why do we use `\z` instead of `\Z`? (ref: https://ruby-doc.org/core-2.3.1/Regexp.html#class-Regexp-label-Anchors)

It wouldn't make much difference, since:
```
>> require 'bigdecimal'
=> true
>> BigDecimal.new("123.45\n")
=> #<BigDecimal:7f89b2185ef0,'0.12345E3',18(18)>
>> BigDecimal.new("123.45\n67")
=> #<BigDecimal:7f89b21be908,'0.12345E3',18(18)>
```
But still, why consider a new line?

filipebarcos: Also, why are you dropping `freeze`?

elfassy: @filipebarcos freezing has no effect on immutable objects (http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/RedundantFreeze)

I'm not a fan of 
```ruby
> BigDecimal.new("123.45\n67").to_f
=> 123.45
```
In my opinion there's too much going on here to justify this being considered as a number. This is similar to `123.45abc` which is also accepted by big decimal but not acceptable for money:
```
> BigDecimal.new("123.45abc").to_f
=> 123.45
```

PR #122, opened 2017-10-10
exclude the money gem files from the deprecation call stack

elfassy:
# Why
Getting a deprecation warning that points back to the money gem is not useful. Since Rails only shows a single line (by default the before last event `caller_locations(2)`) we need to make sure that location is outside of the Money gem.

# What
Exclude all the money gem locations from the callstack sent to the deprecator `warn` method https://github.com/rails/rails/blob/master/activesupport/lib/active_support/deprecation/reporting.rb#L18

### DIFF
--- a/lib/money/deprecations.rb
+++ b/lib/money/deprecations.rb
@@ -7,7 +7,10 @@ Money.class_eval do
 
   def self.deprecate(message)
     if ACTIVE_SUPPORT_DEFINED
-      active_support_deprecator.warn("[Shopify/Money] #{message}\n")
+      external_callstack = caller_locations.reject do |location|
+        location.to_s.include?('gems/money')
+      end
+      active_support_deprecator.warn("[Shopify/Money] #{message}\n", external_callstack)
     else
       Kernel.warn("DEPRECATION WARNING: [Shopify/Money] #{message}\n")
     end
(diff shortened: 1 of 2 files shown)


lib/money/deprecations.rb
```diff
@@ -7,7 +7,10 @@ def self.active_support_deprecator
 
   def self.deprecate(message)
     if ACTIVE_SUPPORT_DEFINED
-      active_support_deprecator.warn("[Shopify/Money] #{message}\n")
+      external_callstack = caller_locations.reject do |location|
+        location.to_s.include?('gems/money')
```
Edouard-chin: I'm a bit worried that this could produce a huge amount of logs

elfassy: Rails will only take the first item from the caller_locations array. This is similar to the rails default `caller_locations(2)` which just sets the start at the second element (http://www.rubydoc.info/stdlib/core/2.0.0/Kernel%3Acaller_locations) 
```
[1] pry(Money)> caller_locations.size
=> 115
[2] pry(Money)> caller_locations.first
=> "/Users/michaelelfassy/.gem/ruby/2.3.3/gems/pry-0.10.4/lib/pry/pry_instance.rb:355:in `eval'"
[3] pry(Money)> caller_locations(2).size
=> 114
[4] pry(Money)> caller_locations(2).first
=> "/Users/michaelelfassy/.gem/ruby/2.3.3/gems/pry-0.10.4/lib/pry/pry_instance.rb:355:in `evaluate_ruby'"
```

elfassy:
🎩  👍

PR #124, opened 2017-10-20
Add MissingCurrency that always raises to force using a real currency

bdewater:
Less magic, more explosions 💥  by setting this as the default, people must pass currency as the second argument to `Money.new` or use a `Money.with_currency` block.

Next step after this PR will be to make it the default of the gem and adjust tests to cover three major currencies (CAD, JPY, JOD) in every spec so we're sure we've covered all common exponents (2, 0, 3 respectively), as discussed IRL.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -11,7 +11,7 @@ class Money
     attr_accessor :parser, :default_currency
 
     def new(value = 0, currency = nil)
-      currency ||= current_currency || default_currency
+      currency ||= resolve_currency
 
       if value == 0
         @@zero_money ||= {}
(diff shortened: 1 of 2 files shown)


elfassy:
> Less magic, more explosions

Lol!

elfassy:
I think i'd much rather if the default_currency is not set to raise, rather than setting a default currency that is a `MissingCurrency`. Right now we initialize the currency before using it so this would raise right away
```ruby
Money.default_currency = MissingCurrency.new #=> raise...
```

It could be instead
```ruby
  currency ||= current_currency || default_currency || raise ArgumentError
```

bdewater:
I like that. I put it in a separate method so we can more easily hook into the behaviour in consuming apps, say prepend a module for shitlists/StatsD metrics.

PR #125, opened 2017-10-30
Money parser refactor

### LINKED ISSUE #13: Parsing is incorrect if there are more than two numbers after the decimal point

elfassy:
# Why
We need a money parser that is currency aware. Passing the currency will now resolve issues with 3 digit currencies.

```ruby
# Before
MoneyParser.parse('1.000') #=> 1000
"1.11111111".to_money.to_f #=> 111111111

# After
MoneyParser.parse('1.000') #=> 1000 (legacy support)
MoneyParser.parse('1.000', 'JOD') #=> 1
MoneyParser.parse('1.000', 'EUR') #=> 1000
"1.11111111".to_money.to_f #=> 1.11
```
fixes #13
Re https://github.com/Shopify/shopify/issues/27704
Re https://github.com/Shopify/shopify/issues/24137
Re https://github.com/Shopify/shopify/issues/134834

# What 
- refactored MoneyParser
- couple of new tests for edge cases

### DIFF
--- a/lib/money/accounting_money_parser.rb
+++ b/lib/money/accounting_money_parser.rb
@@ -1,7 +1,7 @@
 class AccountingMoneyParser < MoneyParser
   private
-  def extract_money(input)
+  def extract_money(input, currency = nil)
     # set () to mean negativity. ignore $
-    super(input.gsub(/\(\$?(.*?)\)/, '-\1'))
+    super(input.gsub(/\(\$?(.*?)\)/, '-\1'), currency)
   end
 end
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -1,31 +1,65 @@
+# Parse an amount from a string
 class MoneyParser
-  # parse a amount from a string
+  MARKS = "\.,·’"
+  EXTRA_MARKS = "\s˙'"
+
   def self.parse(input, currency = nil)
     new.parse(input, currency)
   end
 
   def parse(input, currency = nil)
-    Money.new(extract_money(input.to_s), currency)
+    amount = extract_money(input.to_s, currency)
+    Money.new(amount, currency)
   end
 
   private
-  def extract_money(input)
-    return Money.zero if input.to_s.empty?
 
-    amount = input.scan(/\-?[\d\.\,]+/).first
+  def extract_money(input, currency)
+    return '0' if input.empty?
 
-    return Money.zero if amount.nil?
+    amount = input.scan(/(-?[\d#{MARKS}][\d#{MARKS}#{EXTRA_MARKS}]*)/).first
+    return '0' unless amount
+    amount = amount.first.tr(EXTRA_MARKS, '')
 
-    # Convert 0.123 or 0,123 into what will be parsed as a decimal amount 0.12 or 0.13
-    amount.gsub!(/^(-)?(0[,.]\d\d)\d+$/, '\1\2')
+    *other_marks, last_mark = amount.scan(/[#{MARKS}]/)
+    return amount unless last_mark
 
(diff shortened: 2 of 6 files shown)


lib/money/accounting_money_parser.rb
```diff
@@ -1,7 +1,7 @@
 class AccountingMoneyParser < MoneyParser
   private
-  def extract_money(input)
+  def extract_money(input, currency = nil)
```
sunblaze: Should this argument be `currency_iso` to line up with the method in `MoneyParser`?

lib/money/money_parser.rb
```diff
@@ -1,31 +1,41 @@
...
+  def extract_money(input, currency_iso = nil)
+    return '0' if input.to_s.empty?
+
+    amount = input.scan(/(-?[\d#{MARKS}][\d#{MARKS}#{EXTRA_MARKS}]*)/).first
+    return '0' unless amount
+    amount = amount.first.tr(EXTRA_MARKS, '')
 
-    return Money.zero if amount.nil?
+    *other_marks, last_mark = amount.scan(/[#{MARKS}]/)
+    other_marks.uniq!
```
sunblaze: Why? `other_marks` doesn't seem to be used in this method

lib/money/money_parser.rb
```diff
@@ -1,31 +1,41 @@
...
-    # Convert 0.123 or 0,123 into what will be parsed as a decimal amount 0.12 or 0.13
-    amount.gsub!(/^(-)?(0[,.]\d\d)\d+$/, '\1\2')
+    return amount if last_mark.nil?
 
-    segments = amount.scan(/^(.*?)(?:[\.\,](\d{1,2}))?$/).first
+    *dollars, cents = amount.split(last_mark)
+    dollars = dollars.join.tr(MARKS, '')
 
-    return Money.zero if segments.empty?
+    if cents.size > 2
```
sunblaze: Shouldn't we be checking against the currency's number of subunits?

lib/money/money_parser.rb
elfassy: ah yes, i removed the code that was using this (but will add it back). The idea was that if the other marks are equal to the last mark, then it's for sure a thousand separator. This helps in cases where the currency is not specified. Example `1.000.000`

lib/money/money_parser.rb
elfassy: This `if` is to check if the cents are actually thousands. No currency represents thousands with less than 3 digits, even currencies with 0 or 3 minor units. In other words it's not possible is the cents has less than 3 digits that it's a thousand amount.

lib/money/accounting_money_parser.rb
elfassy: modified `currency_iso` since this also works with a currency object

elfassy:
Doing more 🎩  on the gem bump PR before merging this

elfassy:
updated to keep legacy support `MoneyParser.parse('1.000') #=> 1000 (legacy support)` when no currency is passed. When dealing with a 3 digit currency we just need to pass the currency to get the correct behavior. Europe is a bigger market which may depend on this legacy behavior

PR #127, opened 2017-11-01
Avoid eagerly calculating subunits when it often isn't used.

dylanahsmith:
## Problem

We are spending a good deal of time during sales calculations in GC and a significant number of the object allocations happen within `Money.new`.  For instance, any arithmetic creates a new money object, but only uses the value for the calculation.

## Solution

Having Money#subunits calculate the value so the calculation isn't performed when it isn't necessary.

I was going to have it memoize the result, but we freeze the Money object on initialization which prevents memoization.  However, I don't think that will cause a problem, since we made the same change with `cents` when it was more heavily used.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -4,7 +4,7 @@ class Money
 
   NULL_CURRENCY = NullCurrency.new.freeze
 
-  attr_reader :value, :subunits, :currency
+  attr_reader :value, :currency
   def_delegators :@value, :zero?, :nonzero?, :positive?, :negative?, :to_i, :to_f, :hash
 
   class << self
@@ -88,7 +88,6 @@ class Money
     raise ArgumentError if value.nan?
     @currency = Helpers.value_to_currency(currency)
     @value = value.round(@currency.minor_units)
-    @subunits = (@value * @currency.subunit_to_unit).to_i
     freeze
   end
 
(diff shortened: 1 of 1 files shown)


dylanahsmith:
cc @airhorns

airhorns:
I was gonna suggest unfreezing to set the memo, but googled a bit and it looks like you have to resort to really shitty platform specific hacks to unfreeze, so sounds like thats a bad idea. Is there such a thing as "userland" freezing such that the object could internally support memoization but still present a frozen interface to the client code? Might be worth it if there are other memos we could use? 

My suspicion is no (or that its just too complicated to justify a few memos), such that this PR LGTM.

elfassy:
I'm still not sure why we need to freeze money objects. `attr_reader :value, :currency` so there's no writer method. It does help prevent us from modifying these values inside the gem, but our tests should catch those errors fairly easily. Removing the freeze requirement would unblock us from memoizing just in time.

dylanahsmith:
I think it would be pretty easy for someone to add a mutating method without breaking existing tests, since they cover existing behaviour.

The memoization of `subunits` seems like a premature optimization.  It isn't used that often and it doesn't seem like it would get called multiple times on the same money object, other than for `Money.zero`.

bdewater:
> I'm still not sure why we need to freeze money objects.

That's pretty normal for value objects and a requirement for [`composed_of`](http://api.rubyonrails.org/v5.0.0/classes/ActiveRecord/Aggregations/ClassMethods.html) until we decided to abandon that approach for improving `money_column`.

PR #128, opened 2017-11-06
stricter money parser

elfassy:
# Why
invalid input into the money parser should raise instead of returning 0$
Blocking Shopify/shopify#136489
Re https://github.com/Shopify/stronger_parameters/pull/43

# What
- make this optional (for now) by making this behavior opt-in with a `strict` option
- strict will raise on non numeric strings
- strict will raise on badly formatted thousand seperators

### DIFF
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -1,42 +1,123 @@
+# frozen_string_literal: true
+
 # Parse an amount from a string
 class MoneyParser
-  MARKS = "\.,·’"
-  EXTRA_MARKS = "\s˙'"
+  class MoneyFormatError < ArgumentError; end
 
-  def self.parse(input, currency = nil)
-    new.parse(input, currency)
+  MARKS = %w[. , · ’ ˙ '] + [' ']
+
+  ESCAPED_MARKS = Regexp.escape(MARKS.join)
+  ESCAPED_NON_SPACE_MARKS = Regexp.escape((MARKS - [' ']).join)
+  ESCAPED_NON_DOT_MARKS = Regexp.escape((MARKS - ['.']).join)
+  ESCAPED_NON_COMMA_MARKS = Regexp.escape((MARKS - [',']).join)
+
+  NUMERIC_REGEX = /(
+    [\+\-]?
+    [\d#{ESCAPED_NON_SPACE_MARKS}][\d#{ESCAPED_MARKS}]*
+  )/ix
+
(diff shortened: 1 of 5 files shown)


lib/money/money_parser.rb
```diff
@@ -1,56 +1,140 @@
...
+    [+\-]?
+    (?:
+      (?:\d+)
+      (?:\,\d{4})+
+      (?:.\d+)?
+    )
+  \z/ix
+
+  def self.parse(input, currency = nil, options = {})
+    new.parse(input, currency, options)
```
sunblaze: I don't think this works, named arguments are not interchangeable with a hash at the end of a method. Try adding a test that uses this class method to know for sure

elfassy:
```ruby
[1] pry(main)> class Foo
[1] pry(main)*   def bar(options = {})
[1] pry(main)*     barz(options)
[1] pry(main)*   end
[1] pry(main)*   def barz(a:, b:)
[1] pry(main)*     p [a, b]
[1] pry(main)*   end
[1] pry(main)* end
=> :barz
[2] pry(main)> Foo.new.bar(a: 'it', b: 'works')
["it", "works"]
=> ["it", "works"]
```
I usually prefer this when the method is delegating options. It's also possible to use the `**options` since ruby 2.3. Will add a test that passes an option 👍

lib/money/money_parser.rb
```diff
@@ -1,56 +1,140 @@
...
+      return amount.tr(ESCAPED_NON_DOT_MARKS, '')
+    end
+
+    if amount = number[COMMA_DECIMAL_REGEX]
+      return amount.tr(ESCAPED_NON_COMMA_MARKS, '').sub(',', '.')
+    end
+
+    raise MoneyFormatError, "invalid money string: #{input}" if strict
+
+    Money.deprecate("invalid money strings will raise in the next major release \"#{input}\"")
```
sunblaze: Should we not show this deprecate message if strict is `true`?

elfassy: We only show this message if we don't raise the `MoneyFormatError`

sunblaze: But it's already raising an exception if they're passing `strict` true. Isn't that already proving that they're ready to deal with the change in the next major release?

sunblaze: Nevermind, forgot this method is returning early on successes, not on failures.

PR #131, opened 2017-11-08
Use string column names for money columns and money currency columns

airhorns:
We use Rails' `read_attribute` and `write_attribute` methods to implement money_column stuff, which, in Shopify, is mostly how we get Money objects to work with. In performance critical money-manipulation code, the money column readers and writers actually get invoked a lot, so, we should make them reasonably fast. It turns out that inside `read_attribute` and `write_attribute`, Rails does a lot of string manipulation to check for aliases and index into the actual attributes storage backend. If we pass in a symbol to those methods, the `#to_s` calls that get invoked super often actually end up allocating a remarkable number of strings. If instead we pass strings right into `read_attribute` and `write_attribute`, all those `#to_s` calls are no-ops, so a lot less string garbage is created. So, this change switches the metaprogramming that defines money column methods (and eventually that calls the Rails methods in question) to define using strings instead of symbols, so strings get passed down. It worked! In the Shopify sales ETL, I saw a 60% reduction of immediately garbage collected strings having to do with money manipulation after this change.

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -37,11 +37,13 @@ module MoneyColumn
         else
           raise ArgumentError, 'must set one of :currency_column or :currency options'
         end
+        currency_column = currency_column.to_s.freeze
 
         columns.flatten.each do |column|
-          attribute(column.to_s, MoneyColumn::ActiveRecordType.new)
-          money_column_reader(column, currency_column, currency_iso, coerce_null)
-          money_column_writer(column, currency_column, currency_iso, currency_read_only)
+          column_string = column.to_s.freeze
+          attribute(column_string, MoneyColumn::ActiveRecordType.new)
+          money_column_reader(column_string, currency_column, currency_iso, coerce_null)
+          money_column_writer(column_string, currency_column, currency_iso, currency_read_only)
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/money_column/active_record_hooks.rb
```diff
@@ -37,11 +37,13 @@ def money_column(*columns, currency_column: nil, currency: nil, currency_read_on
         else
           raise ArgumentError, 'must set one of :currency_column or :currency options'
         end
+        currency_column = currency_column.to_s.freeze
```
elfassy: nit: use `# frozen_string_literal: true` at the top of the file instead of `.freeze`

lib/money_column/active_record_hooks.rb
```diff
@@ -37,11 +37,13 @@ def money_column(*columns, currency_column: nil, currency: nil, currency_read_on
...
 
         columns.flatten.each do |column|
-          attribute(column.to_s, MoneyColumn::ActiveRecordType.new)
-          money_column_reader(column, currency_column, currency_iso, coerce_null)
-          money_column_writer(column, currency_column, currency_iso, currency_read_only)
+          column_string = column.to_s.freeze
```
elfassy: same here for `.freeze`

airhorns:
I'm just gonna make sure that Shopify is happy with this change before merging :)

airhorns:
https://buildkite.com/shopify/shopify-branches/builds/384754#05413ab6-fd2c-479c-ad1c-ac1413b0283a looks like it worked! Horray!

PR #132, opened 2017-11-08
bigdecimal 1.3.2

elfassy:
# Why
the new big decimal is more strict as to what string values it supports. This commit makes sure nothing breaks

# What
- support for `Money.new('invalid') #=> 0` with a deprecation warning
- support for `Money.new('1.') #=> 1`

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -2,9 +2,9 @@ require 'bigdecimal'
 
 class Money
   module Helpers
-    extend self
+    module_function
 
-    NUMERIC_REGEX = /\A\s*-?\d*\.?\d*\s*\z/
+    NUMERIC_REGEX = /\A\s*[\+\-]?\d*(\.\d*)?\s*\z/
     DECIMAL_ZERO = BigDecimal.new(0).freeze
 
     def value_to_decimal(num)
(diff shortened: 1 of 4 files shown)


lib/money/helpers.rb
```diff
@@ -23,8 +23,10 @@ def value_to_decimal(num)
         when String
           if num !~ NUMERIC_REGEX
             Money.deprecate("using Money.new('#{num}') is deprecated and will raise an ArgumentError in the next major release")
+            BigDecimal.new(num.to_f, Float::DIG)
```
bdewater: `num.to_d` would do the same?

filipebarcos: Not sure if we care about using this gem without Rails or ActiveSupport (according to gemspec, Rails is just a dev dependency), but plain Ruby should not have `.to_d` right?

bdewater: It's added by requiring `bigdecimal/util` and is in Ruby itself :)

elfassy: we don't require `bigdecimal/util` in this gem, no need for it since it'd be used only once.

My assumption here is that `to_d` behaves the same as `BigDecimal.new` which will raise on invalid values. `to_f` on the other hand "correctly" handles non-numeric `'invalid'` and `'123abc'` values.

elfassy: did some testing and @bdewater is correct, `to_d` doesn't behave like `BigDecimal.new` and will handle invalid values TIL! BUT
```ruby
'1.'.to_d #=> 0
```
🤦‍♂️ 
I'd rather re-implement https://github.com/ruby/bigdecimal/blob/master/lib/bigdecimal/util.rb#L67-L72 here if that's ok then require extra monkey patches

bdewater: I would've expected it to return the same result as `to_f` and it turns out we're not the only one: https://github.com/ruby/bigdecimal/issues/63

PR #135, opened 2017-11-14
Test ensuring currency is preserved when using money_accessor

ptolts:
I think `money_accessor` should probably preserve the currency if you're passing it a `Money` object.


bdewater:
Closing this as https://github.com/Shopify/money/pull/224 has been merged.

PR #138, opened 2017-12-14
have to_json and as_json return amount and currency json

### LINKED ISSUE #76: to_json, as_json and from_json should behave correctly
According to [the Rails docs](http://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html):
`as_json` should return a hash of attributes (amount, currency) to be serialized, not a string
`from_json` is missing to assign attributes from a JSON object

We should not override to_json. A great explanation of the difference between as/to_json can be found here: https://github.com/ohler55/oj/issues/199#issuecomment-61465250

elfassy:
# Why
Fixes #76

# What
- `as_json` should return a hash of attributes (amount, currency) to be serialized, not a string
- `from_json` is missing to assign attributes from a JSON object

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -214,11 +214,11 @@ class Money
   end
 
   def to_json(options = {})
-    to_s
+    JSON.dump({value: value, currency: currency})
   end
 
   def as_json(*args)
-    to_s
+    to_json
   end
 
   def abs
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -32,6 +32,11 @@ class Money
       parser.parse(*args)
     end
 
+    def from_json(json)
+      data = JSON.parse(json)
+      Money.new(data['value'], data['currency'])
+    end
+
     def from_cents(cents, currency = nil)
       new(cents.round.to_f / 100, currency)
     end
@@ -213,12 +218,8 @@ class Money
     cents
   end
 
-  def to_json(options = {})
-    JSON.dump({value: value, currency: currency})
-  end
-
-  def as_json(*args)
-    to_json
+  def as_json(*_args)
+    { value: to_s(:amount), currency: currency.to_s }
   end
 
   def abs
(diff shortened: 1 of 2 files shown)


elfassy:
closing for now in favor or completely removing these methods.

lib/money/money.rb
```diff
@@ -211,11 +216,11 @@ def to_liquid
   end
 
   def to_json(options = {})
-    to_s
+    JSON.dump({value: value, currency: currency})
   end
 
   def as_json(*args)
-    to_s
+    to_json
```
bdewater: This should have the actual implementation.

lib/money/money.rb
```diff
@@ -211,11 +216,11 @@ def to_liquid
   end
 
   def to_json(options = {})
-    to_s
+    JSON.dump({value: value, currency: currency})
```
bdewater: This method should be removed.

lib/money/money.rb
elfassy: this was always wrong too, `as_json` needs to return a hash, not a string... fixed now

elfassy:
DO NOT MERGE... until background jobs can handle the new format

elfassy:
this is ready to merge now since https://github.com/Shopify/shopify/pull/151350 
@bdewater @sunblaze  let me know if you see any other objections. Will run tests in core before merging

lib/money/money.rb
```diff
@@ -213,12 +218,8 @@ def to_liquid
...
   end
 
-  def to_json(options = {})
-    to_s
-  end
-
-  def as_json(*args)
-    to_s
+  def as_json(*_args)
+    { value: to_s(:amount), currency: currency.to_s }
```
sunblaze: Should we prefer `currency.iso_code` here instead to be a bit more clear and future proof if we want to_s to be more human readable?

elfassy:
https://buildkite.com/shopify/shopify-branches/builds/487461#1596f5db-450d-44ed-b569-1a7e2f7ddcbf no go...

sunblaze:
😮

elfassy:
this is the new default behaviour in v1.0, with an option to fallback to the legacy behaviour

PR #140, opened 2018-01-24
parse money strings inside a with_currency block

### LINKED ISSUE #139: Parsing ignores currency provided by with_currency block
```
[1] pry(main)> Money.with_currency('KWD') { Money.parse("100.000") }
=> #<Money value:100000.000 currency:KWD>
[2] pry(main)> Money.parse("100.000")
=> #<Money value:100000.00 currency:>
[3] pry(main)> Money.parse("100.000", 'KWD')
=> #<Money value:100.000 currency:KWD>
```

cc @bdewater

elfassy:
# Why
fixes #139 
right now the currency passed to `with_currency` is not respected
```
> Money.with_currency('KWD') { Money.parse("100.000") }
=> #<Money value:100000.000 currency:KWD> #BUG

> Money.parse("100.000", 'KWD')
=> #<Money value:100.000 currency:KWD>
```

# What 
- properly handle parsing a money string within a `with_currency` block
- `value_to_currency` helper now handles looking up the current/default currency
- some `MoneyParser` consistency cleanup
- fix bug in tests where the `default_currency` was not reset correctly
- removed `Money::Helper.no_currency?(currency)` helper since it's no longer used (`money.no_currency?` is still available)

### DIFF
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -22,7 +22,7 @@ class MoneyParser
     (?:
       (?:\d+)
       (?:[#{ESCAPED_NON_DOT_MARKS}]\d{3})+
-      (?:\.\d+)?
+      (?:\.\d{2,})?
     )
   \z/ix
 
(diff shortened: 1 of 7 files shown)


lib/money/money_parser.rb
```diff
@@ -22,7 +22,7 @@ class MoneyFormatError < ArgumentError; end
     (?:
       (?:\d+)
       (?:[#{ESCAPED_NON_DOT_MARKS}]\d{3})+
-      (?:\.\d+)?
+      (?:\.\d{2,})?
```
filipebarcos: we only support 2 decimals? Example currencies that have 3 decimals: `BHD`, `IQD`,...

elfassy: `{2,}` is 2+

filipebarcos: how bad can I be reading a regex that I don't write. damn me. you're right

PR #141, opened 2018-01-25
rename gem shopify-money

elfassy:
# Why
We want to push this gem to rubygems. 
Because there is already a `money` gem we need to rename ours. 

# What
- rename gem to `shopify-money`
- update docs to show that now the gem needs to be required with option `require: 'money'`
- bump version to `0.10`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ money_column expects a decimal 8,3 database field.
 
 ## Installation
 
-    gem 'money', github: 'shopify/money'
+    gem 'shopify-money', require: 'money'
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ money_column expects a decimal 8,3 database field.
 
 ## Installation
 
-    gem 'money', github: 'shopify/money'
+    gem 'shopify-money', require: 'money'
```
rafaelfranca: if you want you can even rename the file names and constant and just do a `Money = ShopifyMoney` in `lib/shopify-money.rb`. But not required to do in this PR.

PR #142, opened 2018-01-30
parse to use minor units

elfassy:
# Why 
We want to avoid parsing using the currency information as much as possible (since locale is better). In cases where we're not faced with a 3 digit currency, it's false to assume that the thousand delimiter is that of the currency. 

# What 
This PR will make sure we're dealing with a 3 digit currency when checking if the last mark matches the currency's decimal mark.

### DIFF
--- a/spec/money_parser_spec.rb
+++ b/spec/money_parser_spec.rb
@@ -224,6 +224,45 @@ RSpec.describe MoneyParser do
     end
   end
 
+  describe "3 digit decimal currency" do
+    it "parses thousands correctly" do
+      expect(@parser.parse("1,111", "JOD")).to eq(Money.new(1_111, 'JOD'))
+      expect(@parser.parse("1.111.111", "JOD")).to eq(Money.new(1_111_111, 'JOD'))
+      expect(@parser.parse("1 111", "JOD")).to eq(Money.new(1_111, 'JOD'))
+    end
+
+    it "parses decimals correctly" do
+      expect(@parser.parse("1.111", "JOD")).to eq(Money.new(1.111, 'JOD'))
+      expect(@parser.parse("1,11", "JOD")).to eq(Money.new(1.11, 'JOD'))
+    end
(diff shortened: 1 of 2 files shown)


spec/money_parser_spec.rb
```diff
@@ -224,6 +224,45 @@
     end
   end
 
+  describe "3 digit decimal currency" do
+    it "parses thousands correctly" do
+      expect(@parser.parse("1,111", "JOD")).to eq(Money.new(1_111, 'JOD'))
+      expect(@parser.parse("1.111.111", "JOD")).to eq(Money.new(1_111_111, 'JOD'))
```
clauderic: Currently `"1.111.111"` doesn't work like that with the JS parser for JOD, I'll need to fix that

spec/money_parser_spec.rb
```diff
@@ -224,6 +224,45 @@
...
+    it "parses decimals correctly" do
+      expect(@parser.parse("1,11", "JPY")).to eq(Money.new(1, 'JPY'))
+      expect(@parser.parse("1.11", "JPY")).to eq(Money.new(1, 'JPY'))
+    end
+  end
+
+  describe "two decimal currency" do
+    it "parses thousands correctly" do
+      expect(@parser.parse("1,111", "USD")).to eq(Money.new(1_111, 'USD'))
+      expect(@parser.parse("1.111", "USD")).to eq(Money.new(1_111, 'USD'))
```
clauderic: Currently since `.` is the decimal mark for USD this returns `1.11` with the current JS parser. Not sure which one should be the expected behaviour

elfassy: yea that's a tough one. This is the legacy behavior of the money gem to handle europeans entering amounts with dot

joshnuss:
Wish we would have currency formatting settings in the admin, that way we won't have to guess what the user wants.

PR #145, opened 2018-04-02
add bundle gem tasks for shipit

elfassy:
https://shipit.shopify.io/shopify/money/production/deploys/546605
```
$ bundle exec rake release
pid: 26413
rake aborted!
Don't know how to build task 'release' (see --tasks)
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,11 @@
 # encoding: utf-8
 
+begin
+  require 'bundler/gem_tasks'
+rescue LoadError
+  nil
+end
+
 require 'rubygems'
 require 'bundler'
 require 'rdoc/task'


Rakefile
```diff
@@ -1,6 +1,12 @@
 # encoding: utf-8
 
-require 'rubygems'
+begin
+  require 'bundler/setup'
+  require 'bundler/gem_tasks'
```
bdewater: AFAIK this is the only one we need, looking at the Rakefile from a freshly generated gem. Template here: https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/Rakefile.tt

elfassy: `bundler/setup` lets you skip the whole `Bundler.setup` block but since we already have that, just removed it. Should be good now

elfassy:
ping @bdewater @filipebarcos are we good with this PR?

PR #146, opened 2018-04-03
Fix float rounding issue

arturmalecki:
WIP


lib/money/helpers.rb
```diff
@@ -7,6 +7,7 @@ module Helpers
 
     NUMERIC_REGEX = /\A\s*[\+\-]?\d*(\.\d+)?\s*\z/
     DECIMAL_ZERO = BigDecimal.new(0).freeze
+    MAX_DECIMAL = 20
```
elfassy: we should go for 21 since there was talk of modifying the DB to add an extra decimal (and be `(21,3)`)

spec/money_spec.rb
```diff
@@ -241,6 +241,10 @@
     expect((0.33 * Money.new(0.10))).to eq(Money.new(0.03))
   end
 
+  it "handles multiplication for larg numbers" do
+    expect((Money.new("999999999999999999.99") * -1).value).to eql(BigDecimal.new("-999999999999999999.99", 20))
```
elfassy: that's nine hundred ninety-nine quadrillion!!! :) 
https://www.calculatorsoup.com/calculators/conversions/numberstowords.php

arturmalecki: Yes, this is. And it was in `tax_amout` 💰

PR #155, opened 2018-05-14
to_money assign currency when none

elfassy:
# Why
Money with no currency should always take the provided currency when one is available.

# What 
```ruby
money = Money.new(1)
money.to_money('USD')
money.currency.to_s == 'USD' #=> true
```

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -183,7 +183,16 @@ class Money
     [ReverseOperationProxy.new(other), self]
   end
 
-  def to_money(_currency = nil)
+  def to_money(curr = nil)
+    if !curr.nil? && no_currency?
+      return Money.new(value, curr)
+    end
+
+    curr = Helpers.value_to_currency(curr)
+    unless currency.compatible?(curr)
+      Money.deprecate("mathematical operation not permitted for Money objects with different currencies #{curr} and #{currency}.")
+    end
+
     self
   end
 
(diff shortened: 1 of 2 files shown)


spec/money_spec.rb
```diff
@@ -38,6 +38,10 @@
     expect(money.to_money).to eq(money)
   end
 
+  it "#to_money uses the provided currency when it doesn't already have one" do
```
zacharyreynochiasson: Maybe add test case for `to_money` on `Money` with currency to verify that it doesn't override?

elfassy: good call, done

spec/money_spec.rb
```diff
@@ -38,6 +38,14 @@
...
+    expect(Money.new(1).to_money('CAD')).to eq(Money.new(1, 'CAD'))
+  end
+
+  it "#to_money doesn't overwrite the currency" do
+    expect(Money.new(1, 'USD').to_money('CAD')).to eq(Money.new(1, 'USD'))
```
bdewater: This might become a source of hard to debug bugs. Should we mark this behaviour deprecated?

elsom25: I also wonder if we want to warn (kinda like we do with `MoneyBag` anytime that currency param differs from the underlying currency?)

lib/money/money.rb
```diff
@@ -183,7 +183,11 @@ def coerce(other)
     [ReverseOperationProxy.new(other), self]
   end
 
-  def to_money(_currency = nil)
+  def to_money(curr = nil)
+    if !curr.nil? && no_currency?
+      return Money.new(value, curr)
+    end
+
```
elsom25:
```rb
if curr != currency
  # deprecate
end

self
```
?

PR #156, opened 2018-05-23
Allow Money.zero to specify a currency for strictness

elsom25:
Sometimes we want a specific currency around our zero money instead of a null currency

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -23,8 +23,8 @@ class Money
     end
     alias_method :from_amount, :new
 
-    def zero
-      new(0, NULL_CURRENCY)
+    def zero(currency = NULL_CURRENCY)
+      new(0, currency)
     end
     alias_method :empty, :zero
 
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -23,8 +23,9 @@ def new(value = 0, currency = nil)
     end
     alias_method :from_amount, :new
 
-    def zero
-      new(0, NULL_CURRENCY)
+    def zero(currency = nil)
+      currency = Helpers.value_to_currency(currency) if currency.present?
+      new(0, currency.presence || NULL_CURRENCY)
```
grcooper: why `currency.presence` and not iso? I'm just not well versed in what a currency is and couldn't find presence in the currency model

elsom25: [`#presence`](http://guides.rubyonrails.org/active_support_core_extensions.html#presence) is from ActiveSupport and basically coerces empty values to nil. We could probably do the simple `currency || NULL_CURRENCY`, but I like using presence to protect against weird edges (i.e. `.zero("")`

elfassy: this should do it:
```ruby
def zero(currency = NULL_CURRENCY)
  new(0, currency)
end
```

elsom25: 🤦‍♂️

PR #163, opened 2018-08-14
Add IncompatibleCurrencyError

elfassy:
# What

Add a `Money::IncompatibleCurrencyError`

# Why

This error is not used yet, but it will help people get prepared for the next release


lib/money/errors.rb
```diff
@@ -0,0 +1,9 @@
+class Money
+  class Error < StandardError
+  end
+
+  class IncompatibleCurrencyError < Error
+  end
+end
+
```
elsom25: nit: extra newline

PR #165, opened 2018-08-14
parser delegate non string to money

elfassy:
# Why
Avoid parsing non string inputs
For now we want to avoid a scenario like 8% tax 
```ruby
Money.parse(1.66 * 0.8) #=> Money.new(1328), should be Money.new(1.33)
```

# What

- delegate anything passed to `Money.parse` that is not a string to `Money.new`

### DIFF
--- a/lib/money/money_parser.rb
+++ b/lib/money/money_parser.rb
@@ -63,13 +63,17 @@ class MoneyParser
 
   def parse(input, currency = nil, strict: false)
     currency = Money::Helpers.value_to_currency(currency)
-    amount = extract_money(input.to_s, currency, strict)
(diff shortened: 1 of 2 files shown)


bdewater:
Change looks good to me but I don't follow the explanation: our core extension for the Numeric class already calls `Money.new`?
https://github.com/Shopify/money/blob/5df7243f39e2b1ae5f8b36cadc066e8c97e9df48/lib/money/core_extensions.rb#L4-L8

So this is an issue only when people call `Money.parse(1.328)`?

elfassy:
@bdewater yes, just updated the description :D

PR #166, opened 2018-09-17
JSON serialize to a string in both ActiveSupport and JSON

XrXr:
Before this commit, ActiveSupport and the JSON gem serializes
Money objects to different things due to `as_json` and `to_json`
having different interfaces.

This patch makes it so that under both libraries, Money consistently
serializes to a string in JSON.

I found this issue while trying to get core to run on Ruby 2.5.1. I needed to upgrade Oj and the new Oj behaves exactly like the json gem (under our config) while the Oj we currently use behave like ActiveSupport. So this is for getting the current behaviour in the updated Oj. ~I can work around it in core by telling Oj to behave like ActiveSupport again, but I thought I open something here.~

Edit: I actually might need this merged for the update.

Related to #76

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -215,7 +215,7 @@ class Money
   end
 
   def to_json(options = {})
-    to_s
+    "\"#{to_s}\""
   end
 
   def as_json(*args)
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -83,6 +83,10 @@ RSpec.describe "Money" do
     expect(money.as_json).to eq("1.00")
   end
 
+  it "has a consistent json encoding" do
+    expect(ActiveSupport::JSON.encode(money)).to eq(::JSON.dump(money))
+  end
+
   it "is constructable with a BigDecimal" do
     expect(Money.new(BigDecimal("1.23"))).to eq(Money.new(1.23))
   end
@@ -283,7 +287,7 @@ RSpec.describe "Money" do
   end
 
   it "returns cents in to_json" do
-    expect(Money.new(1.00).to_json).to eq("1.00")
+    expect(Money.new(1.00).to_json).to eq('"1.00"')
   end
 
   it "supports absolute value" do


elfassy:
do you have a gem bump PR in core with this sha? I'd like to see that tests pass before merging this. I had attempted to fix the json representation here https://github.com/Shopify/money/pull/138 and ran into many tests expecting incorrect json values.

XrXr:
https://buildkite.com/shopify/shopify/builds/88357
We do get a few failures :(
It looks like core expects Money to JSON serialize to double or string depending on the component, so if I want Money to consistently JSON serialize to *anything*, I can't avoid changing core.   

I also tried the alternative version of making Money always json serialize to a double. That unfortunately results in even more failures.
https://buildkite.com/shopify/shopify/builds/88512

I don't want to keep the current idiosyncrasy, but maybe I will have to look elsewhere to do the Ruby update...

tjoyal:
I'm not sure I'm the best person to review this, but let me share my thoughts and tell me what you think.

So this PR aim at "fixing the json representation of a Money object"?

From what I understand, a `Money` object is a mix of a "value" and a "currency". I am not certain why we aren't aiming to set the serialization to represent the object in it's most native form and aim to make it a string instead. It might require some work, but wouldn't it make things better down the road to have a json representation of this complex object via multiple keys?

lib/money/money.rb
```diff
@@ -215,7 +215,7 @@ def to_liquid
   end
 
   def to_json(options = {})
-    to_s
+    "\"#{to_s}\""
```
tjoyal: If changing the representation at this time is not what we want, and the idea is to keep the core representation as a "string", then I'd say something that would achieve something similar yet provide more meaning to the code deferring the serialization logic to an existing layer that understand how to dump a string as json.

Eg.:
```ruby
def to_json(options = {})
  to_s.to_json
end
```

What worries me here is the options are discarded and I don't know enough about rails serialization to know if this comes from Rails or a param to the method that shouldn't have existed in the first place. It does not prevent the changes proposed here, but raises to me question as to if this new method definition is still broken in some way.

XrXr:
> It might require some work, but wouldn't it make things better down the road to have a json representation of this complex object via multiple keys?

The core issue I'm trying to fix here is consistency. I went with the string representation only because it causes the least amount of disruption.

I'm not sure whether going to an object representation is worth the cost. Judging from the number of failed tests we get, people seem to expect Money to serialize into a number, whether in string form or not. I suspect this is because whenever people JSON serialize, they are interacting with some API over http, and it's nice to be able to just stick Money objects into a hash and tell a JSON lib to serialize that. It might be useful to count the usage to figure out which design is better.

XrXr:
Closing since this is really old and a change like this will probably break a lot of things :)

PR #167, opened 2018-10-16
Enable money_column to write currency to DB when default currency is nil

tomgatzgates:
## What

When `Money`'s default currency is set to `nil`, `money_column`, more specifically `#write_money_attribute`, fails to write the currency provided by the input Money object to the database. Instead it raises `ArgumentError, 'missing currency'`. 

Line 51 is only checking for the currency from one of the following: 
- Money's default currency
- The column's hardcoded currency
- The currency already assigned to the record's corresponding currency column

https://github.com/Shopify/money/blob/df3204ea232a657414f1c5f421d206a50be9dac4/lib/money_column/active_record_hooks.rb#L41-L64

## How

Enable the raw currency source to fall back to the input money object's currency. 
First check the default currency, then the currency already in the column finally, check the input currency. 
Obviously we want to rescue from the input not being a Money object so always default to nil. 

Note: If a `Money.default_currency` is set, options['currency'] will return the default currency so we'll never evaluate the column or input. 

## Why

This enables applications that don't have a default currency set to still write currency to the database with `money_column`.

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -48,19 +48,14 @@ module MoneyColumn
         return self[column] = nil
       end
 
-      currency_raw_source = begin
-        options[:currency] || send(options[:currency_column]) || money.currency
-      rescue
-        nil
-      end
-
-      currency_source = Money::Helpers.value_to_currency(currency_raw_source)
+      currency_raw_source = options[:currency] || (send(options[:currency_column]) rescue nil)
 
       if !money.is_a?(Money)
-        return self[column] = Money.new(money, currency_source).value
+        return self[column] = Money.new(money, currency_raw_source).value
       end
 
-      if currency_raw_source && !currency_source.compatible?(money.currency)
+      currency_source = currency_raw_source ? Money::Helpers.value_to_currency(currency_raw_source) : Money::NULL_CURRENCY
+      unless currency_source.compatible?(money.currency)
         Money.deprecate("[money_column] currency mismatch between #{currency_source} and #{money.currency}.")
       end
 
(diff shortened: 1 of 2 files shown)


lib/money_column/active_record_hooks.rb
```diff
@@ -48,7 +48,12 @@ def write_money_attribute(column, money)
         return self[column] = nil
       end
 
-      currency_raw_source = options[:currency] || (send(options[:currency_column]) rescue nil)
+      currency_raw_source = begin
+        options[:currency] || send(options[:currency_column]) || money.currency
```
bdewater: Unrelated to the order of determining the currency source, it would be nice to clean up this very broad `rescue`. From what I can tell only `NoMethodError` could raise here, and we can avoid leaning on this crutch for controlling flow without too much hassle.

Something like:
```ruby
currency_raw_source = options[:currency] || 
  (send(options[:currency_column]) if respond_to(options[:currency_column])) || 
  money.currency
```
WDYT?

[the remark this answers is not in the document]
tomgatzgates: I think it should - because a couple of lines down we check if the currency passed in is allowed, aka does it match a specified currency already on that column. I.e. default is 'CAD` but you pass in 'GBP', it should fail.

tomgatzgates: That's an assignment with an `OR` followed by a trailing `if` which itself contains an `OR` not to mention using send and respond to just feel funny to me. 
That's some real gymnastic code ha

tomgatzgates:
I'm unsure what the intended behaviour is/should be with those two examples? 
It seems odd to fail hard when trying to write a currency to the db just because a default currency is not set. The fact its passing when a default currency is set is more worrying to me because it's enabling it to pass through line 51 safely.

PR #171, opened 2018-11-06
no deprecation updating non-read-only currency column

elfassy:
# Why
When `currency_read_only` is false or not set, we want to update the currency with that of the money column associated. Since it's the expected behaviour, there should not be a deprecation warning.

# What 
Skip the deprecation warning when `currency_read_only` is false or not set

### DIFF
--- a/spec/money_column_spec.rb
+++ b/spec/money_column_spec.rb
@@ -139,9 +139,8 @@ RSpec.describe 'MoneyColumn' do
     expect(record.currency.to_s).to eq('EUR')
   end
 
-  it 'does overwrite a currency if changed but will show a deprecation notice' do
+  it 'does overwrite a currency' do
     expect(record.currency.to_s).to eq('EUR')
-    expect(Money).to receive(:deprecate).once
     record.update(price: Money.new(4, 'JPY'))
     expect(record.currency.to_s).to eq('JPY')
   end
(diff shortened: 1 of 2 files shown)


spec/money_column_spec.rb
```diff
@@ -139,9 +139,8 @@ class MoneyClassInheritance2 < MoneyWithCustomAccessors
     expect(record.currency.to_s).to eq('EUR')
   end
 
-  it 'does overwrite a currency if changed but will show a deprecation notice' do
+  it 'does overwrite a currency' do
     expect(record.currency.to_s).to eq('EUR')
-    expect(Money).to receive(:deprecate).once
```
bdewater: Would be good for completeness sake to keep this old test around with the new test.

elfassy: not sure what you mean, the old test doesn't pass anymore since we don't want to show the deprecation

bdewater: Sorry, meant the old test could be kept around if we added the `currency_read_only` option to test the deprecation.

elfassy: ah, sorry makes sense. We are already covered by this one: https://github.com/Shopify/money/pull/171/files#diff-1a873d5a205667d2b652832f32bdc7c6R220

PR #172, opened 2018-11-06
Fix cache when using update_columns

elfassy:
When a money column is updated with `update_columns` or `update_column` it will bypass the `write_money_attribute` and write directly to the db and this will make it only save the value and not the currency

workarounds
`.update_attribute(price: money)`
or
call `.reload` after updating the columns

### DIFF
--- a/lib/money_column/active_record_hooks.rb
+++ b/lib/money_column/active_record_hooks.rb
@@ -25,6 +25,11 @@ module MoneyColumn
       super
     end
 
+    def write_attribute_without_type_cast(attr_name, _value, *args)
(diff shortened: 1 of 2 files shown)


tomgatzgates:
When calling the money column attribute, the values get cached into the `money_column_cache`. 

The intended behaviour when using `update_columns` is that the updated column, regardless of whether it's currency or the decimal value, should reflect the changes written to the db. 

```ruby
class MoneyRecord < ActiveRecord::Base
end
```

elfassy:
screen shot 2018-11-06 at 4 29 06 pm

elfassy:
closing for now, i've added some workarounds to the description.

PR #173, opened 2018-11-27
Show the column that's meant in deprecation warning

robinbrandt:
This saves the developer some debugging time to figure out which column is actually meant in the deprecation warning.


bdewater:
Thanks!

PR #175, opened 2019-01-07
Update allocate to assign extra pennies more fairly

sunblaze:
The current implementation favours the order of the splits too highly. This results in unexpected subunits (pennies) being assigned to the first ratios, when the decimal value shows a value higher than half a subunit.

For example.

If you split 5 subunits between two parties with 45% to the first party and 55% to the second. You'd expect the second party to get the extra subunit (55% of 5 is 2.75 vs 45% of 5 is 2.25). But the previous implementation would only floor the amounts then assign the remaining subunits round robin, this resulted in the first party getting the extra subunit which is not usually what's you'd expect from a split.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -254,14 +254,16 @@ class Money
   #     #=> [#<Money value:2.50 currency:USD>, #<Money value:1.25 currency:USD>, #<Money value:1.25 currency:USD>]
   #   Money.new(5, "USD").allocate([0.3, 0.7])
   #     #=> [#<Money value:1.50 currency:USD>, #<Money value:3.50 currency:USD>]
+
+  # @example extra pennies are first distributed to the split that is the closest to rounding up 55% of 0.05 is 0.0275
+  #   Money.new(0.05, "USD").allocate([0.45, 0.55])
+  #     #=> [#<Money value:0.02 currency:USD>, #<Money value:0.03 currency:USD>]
+
+  # @example left over cents are distributed round robin when splits are equal
   #   Money.new(100, "USD").allocate([0.33, 0.33, 0.33])
   #     #=> [#<Money value:33.34 currency:USD>, #<Money value:33.33 currency:USD>, #<Money value:33.33 currency:USD>]
 
-  # @example left over cents distributed to first party due to rounding, and two solutions for a more natural distribution
-  #   Money.new(30, "USD").allocate([0.667, 0.333])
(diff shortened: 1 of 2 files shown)


jahfer:
I understand the principle of this, but I am somewhat confident this is a breaking change that may have big ramifications. The Sales component currently has many areas where we perform allocations and take `n` parts of the split to make sure we know what remains of the current value.

As an example, let's say we're trying to figure out how much remains to refund for an item.

```ruby
# SCENARIO:
#  6 x T-Shirt = $10.00
# -1 x T-Shirt (refunded in the past)
# -3 x T-Shirt (let's compute refund below...)

total_price = Money.new(10, 'USD')
already_refunded = Money.new('1.67', 'USD')
```

Before this PR:
```ruby
total_price.allocate([Rational(1,6), Rational(3,6), Rational(2,6)])
# => [#<Money value:1.67 currency:USD>, #<Money value:5.00 currency:USD>, #<Money value:3.33 currency:USD>]
```

After this PR:
```ruby
total_price.allocate([Rational(1,6), Rational(3,6), Rational(2,6)])
# => [#<Money value:1.66 currency:USD>, #<Money value:5.01 currency:USD>, #<Money value:3.33 currency:USD>]
```

This is problematic because our system would see that we've only refunded $1.66 and still have $8.34 left to refund, when in reality we only have $8.33 left.

In order to be compatible both before and after this change, we would have to do something like this:

```ruby
previously_allocated = already_refunded / total_price # can't do this on Money
current_allocation = Rational(3,6)
remaining = 1 - current_allocation - previously_allocated

# Effectively 1/6, 3/6, 2/6, but we are not relying on previous
# assumptions. This would work fine both before and after our change
# but we don't really write code like this in Core AFAIK.
total_price.allocate([previously_allocated, current_allocation, remaining])
```

sunblaze:
I thought this was the case probably the case. Would adding a `fair_allocate` be preferable so it doesn't break current dependencies? That way we would opt to use it in the future easily in core if we want to switch to remaining amount method?

jahfer:
I don't have anything against introducing a new method like this, but is there a strong use case for this though? It's fairly complex to understand how the Money gem works already, and I'm curious where this problem came from. Was this driven by a merchant complaint?

I don't expect this to have cropped up many times in the gem's lifetime and I don't know that it's worth complicating the API over a minimal tweak like this.

sunblaze:
I ran into this need when I was working with the foreign exchange rate code, but changes in the implementation has made this unnecessary now. I wouldn't want to over complicate matters. Closing for now then

PR #176, opened 2019-01-09
Add new method for calulating splits without needing to return an array of splits.

FranOis:
Returning an array of Money might not always be the best solution when all you care about it to mutate and transform the returned Array. I'm proposing we add a new method that returns the splits information so we can do arithmetics on the information without requiring a new Array. Sometimes returning an array works, but some other times, it causes a lot of stress on the RAM with a very high num parameter.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -340,14 +340,32 @@ class Money
   # @example
   #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def split(num)
+    calculate_splits(num).sum { |value, count| Array.new(count, value) }
+  end
+
+  # Calculate the splits evenly without losing pennies.
+  # Returns the number of high and low splits and the value of the high and low splits.
+  # Where high represents the Money value with the extra penny
+  # and low a Money without the extra penny.
+  #
+  # @param [2] number of parties.
+  #
+  # @return [Hash<Money, Integer>]
+  #
+  # @example
+  #   Money.new(100, "USD").calculate_splits(3) #=> {Money.new(34) => 1, Money.new(33) => 2}
+  def calculate_splits(num)
     raise ArgumentError, "need at least one party" if num < 1
     subunits = self.subunits
     low = Money.from_subunits(subunits / num, currency)
     high = Money.from_subunits(low.subunits + 1, currency)
 
-    remainder = subunits % num
+    num_high = subunits % num
 
-    return Array.new(remainder, high) + Array.new(num - remainder, low)
+    {}.tap do |result|
+      result[high] = num_high if num_high > 0
+      result[low] = num - num_high
+    end
   end
 
   # Clamps the value to be within the specified minimum and maximum. Returns
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -618,6 +618,25 @@ RSpec.describe "Money" do
     end
   end
 
+  describe "calculate_splits" do
+    specify "#calculate_splits gives 1 cent to both people if we start with 2" do
+      actual = Money.new(0.02, 'CAD').calculate_splits(2)
+
+      expect(actual).to eq({
+        Money.new(0.01, 'CAD') => 2,
+      })
+    end
+
+    specify "#calculate_splits gives an extra penny to one" do
+      actual = Money.new(0.04, 'CAD').calculate_splits(3)
+
+      expect(actual).to eq({
+        Money.new(0.02, 'CAD') => 1,
+        Money.new(0.01, 'CAD') => 2,
+      })
+    end
+  end
+
   describe "fraction" do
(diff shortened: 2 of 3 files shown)


elfassy:
fixed failing test here https://github.com/Shopify/money/pull/178

FranOis:
@elfassy awesome! thanks for this. I've rebased and now everything is green :)

spec/money_spec.rb
```diff
@@ -618,6 +618,16 @@
     end
   end
 
+  describe "calculate_splits" do
+    specify "#calculate_splits gives 1 cent to both people if we start with 2" do
+      number_high, number_low, high, low = Money.new(0.02).calculate_splits(2)
```
bdewater: Can you specify a currency for this money object? We're looking to deprecate instantiating Money objects without currency.

spec/money_spec.rb
```diff
@@ -618,6 +618,16 @@
...
+      expect(number_high).to eq(0)
+      expect(number_low).to eq(2)
+      expect(high.value).to eq(0.02)
+      expect(low.value).to eq(0.01)
```
bdewater: Assuming the currency will be CAD, why not `expect(low).to eq(Money.new(0.01, "CAD"))`?

spec/money_spec.rb
FranOis: Done :)

spec/money_spec.rb
FranOis: Done :)

lib/money/money.rb
```diff
@@ -340,14 +340,29 @@ def allocate_max_amounts(maximums)
...
+  # Returns the number of high and low splits and the value of the high and low splits.
+  # Where high represents the Money value with the extra penny
+  # and low a Money without the extra penny.
+  #
+  # @param [2] number of parties.
+  #
+  # @return [Array<Integer, Integer, Money, Money>]
+  #
+  # @example
+  #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
```
elfassy: example needs to be updated for `calculate_splits`

lib/money/money.rb
```diff
@@ -340,14 +340,29 @@ def allocate_max_amounts(maximums)
...
     raise ArgumentError, "need at least one party" if num < 1
     subunits = self.subunits
     low = Money.from_subunits(subunits / num, currency)
     high = Money.from_subunits(low.subunits + 1, currency)
 
-    remainder = subunits % num
-
-    return Array.new(remainder, high) + Array.new(num - remainder, low)
+    num_high = subunits % num
+    [num_high, (num - num_high), high, low]
```
elfassy: what was the argument against 
```ruby
{ high => num_high, low => num_low }
```
the split method would become
```ruby
  def split(num)
    calculate_splits(num).sum { |value, count| Array.new(value, count) }
  end
```

FranOis: As long as you don't care to know which is which it would work.

FranOis:
@bdewater Anything left blocking this PR?

FranOis:
@bdewater ping :)

lib/money/money.rb
```diff
@@ -340,14 +340,28 @@ def allocate_max_amounts(maximums)
...
     raise ArgumentError, "need at least one party" if num < 1
     subunits = self.subunits
     low = Money.from_subunits(subunits / num, currency)
     high = Money.from_subunits(low.subunits + 1, currency)
 
-    remainder = subunits % num
-
-    return Array.new(remainder, high) + Array.new(num - remainder, low)
+    num_high = subunits % num
+    {high => num_high, low => num - num_high}
```
elfassy: I'd suggest removing the amounts with zero events
```suggestion
     {}.tap |result|
       result[high] = num_high if num_high > 0
       result[low] = num - num_high
     end
```

lib/money/money.rb
```diff
@@ -340,14 +340,32 @@ def allocate_max_amounts(maximums)
   # @example
   #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def split(num)
+    calculate_splits(num).sum { |value, count| Array.new(count, value) }
```
rafaelfranca: Are you sure this works? You can sum array using `sum` without argument.

```
>> {1 => 2}.sum { |value, count| Array.new(count, value) }
Traceback (most recent call last):
        4: from /opt/rubies/2.5.3/bin/irb:11:in `<main>'
        3: from (irb):7
        2: from (irb):7:in `sum'
        1: from (irb):7:in `+'
TypeError (Array can't be coerced into Integer)
```

FranOis: Test were passing since in dev we are adding a dependency on Rails which adds ActiveSupport to add it. I've created this PR to fix this: https://github.com/Shopify/money/pull/180

PR #177, opened 2019-01-18
Update MoneyAccessor to keep currency

drose-shopify:
Updates the Money constructor to use
the values currency if no valid currency
is passed in and the value passed in is a Money

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -11,8 +11,13 @@ class Money
     attr_accessor :parser, :default_currency
 
     def new(value = 0, currency = nil)
+      currency = if value.is_a?(Money) && currency.blank?
+        Helpers.value_to_currency(value.currency)
+      else
+        Helpers.value_to_currency(currency)
+      end
+
       value = Helpers.value_to_decimal(value)
-      currency = Helpers.value_to_currency(currency)
 
       if value.zero?
         @@zero_money ||= {}
--- a/spec/money_accessor_spec.rb
+++ b/spec/money_accessor_spec.rb
@@ -69,6 +69,12 @@ RSpec.shared_examples_for "an object with a money accessor" do
 
     expect(object.price).to eq(nil)
   end
+
+  it 'preserves the currency value when passed to money_accessor' do
+    object = described_class.new(Money.new(10, 'USD'))
+    currency = Money::Currency.new('USD')
+    expect(object.price.currency).to eq(currency)
+  end
 end
 
 RSpec.describe NormalObject do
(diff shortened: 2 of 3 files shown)


lib/money/money.rb
```diff
@@ -11,8 +11,13 @@ class << self
     attr_accessor :parser, :default_currency
 
     def new(value = 0, currency = nil)
+      currency = if value.is_a?(Money) && currency.blank?
+        Helpers.value_to_currency(value.currency)
+      else
+        Helpers.value_to_currency(currency)
+      end
```
elfassy: Should we just return the value instead of creating a new object?
```ruby
  return value if value.is_a?(Money)
  ...
```
I'm not sure what is the use case for this though. Seems to me like we don't know if the object is a money or not yet, which should not happen. It can create lots of confusion if the currency don't match
`Money.new(Money.new(1, 'USD'), 'JPY')` 😨 

Either way Money gem doesn't have an explicit dependency on rails, so we can't use `blank?`.  We could do something like
```ruby
if currency.nil? && value.is_a?(Money)
  currency = value.currency
end
currency = Helpers.value_to_currency(currency)
```
But again, not sure this is a valid use case. We might be better off raising if someone passes a Money object.

vignesh-vs-in: Currently initializing money with money throws away currency and only keeps value, which is even more confusing.

Other extended types such as big decimal support initializing with the same type.

elfassy: We can add a deprecation warning to see who's passing in a money object. Assuming there are not many places, i'd rather throw an exception and let people handle these cases higher up.

drose-shopify: I can add that, but it feels weird to me that we will essentially require people to do:
```
value = if maybe_a_money.is_a?(Money)
  maybe_a_money
else
 Money.new(maybe_a_money)
end
```

Just becomes tedious when we could handle this for them with a better API.
I'd also be happy to create a new static method that would be able to handle this case (maybe it takes in a value that could be a hash)

elfassy: `maybe_a_money.to_money('USD')` is what is being used to do this, although it will raise if the currency don't match. I'd like to see a real world example of where we don't know what type the `maybe_a_money` is. I have a feeling there's a problem higher up.

elfassy: ok to unblock this PR let's do
```ruby
  return value.to_money(currency) if value.is_a?(Money)
  ...
```
This will keep the existing behaviour of being able to pass a money object to the constructor but will keep the currency (and show a deprecation/raise `Money::IncompatibleCurrencyError` if the currencies don't match)

PR #180, opened 2019-02-12
Fix sum issue when not running with Rails

FranOis:
Since Rails is not supposed to be a prod dependency for Money gem. Rewrite the `split` method so it doesn't depend on an Active Support feature.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -340,7 +340,7 @@ class Money
   # @example
   #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def split(num)
-    calculate_splits(num).sum { |value, count| Array.new(count, value) }
+    calculate_splits(num).sum([]) { |value, count| Array.new(count, value) }
   end
 
   # Calculate the splits evenly without losing pennies.


lib/money/money.rb
```diff
@@ -340,7 +340,7 @@ def allocate_max_amounts(maximums)
   # @example
   #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def split(num)
-    calculate_splits(num).sum { |value, count| Array.new(count, value) }
+    calculate_splits(num).inject([]) { |acc, split| acc + Array.new(split[1], split[0]) }
```
bdewater: Ruby 2.4 actually added [Enumerable#sum](https://ruby-doc.org/core-2.4.0/Enumerable.html#method-i-sum) but with slightly [different semantics](https://github.com/rails/rails/blob/e9b1b8d2f0cd3899067174b433121344e87eb878/activesupport/lib/active_support/core_ext/enumerable.rb#L4-L9) than Active Support's version. I'm wondering if we could make it work with Ruby's native sum by passing `[]` as the identity argument?

And yeah, we should either make the dependency explicit or fix our builds to test with and without Rails integration 😞

FranOis: TIL: I tested in a separate Ruby file and ran it, the output was an array with 5 10s and 2 8s.

```
a = {
  10 => 5,
  8 => 2
}

b = a.sum([]) { |value, count| Array.new(count, value) }

puts b
```

bdewater:
Could you update the commit message to reflect the new code? After that I believe it's good to go.

PR #181, opened 2019-02-14
Add method sum_splits

FranOis:
Move `calculate_splits` to a private method and add method `sum_splits`.

While less flexible, this new method is clearer for the reader to figure out what it is doing. I've chosen this approach since it is similar to how `Enumerable#sum` works. This gives a familiar interface for any Ruby user.

cc @elsom25 @cpauderis

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -340,32 +340,33 @@ class Money
   # @example
   #   Money.new(100, "USD").split(3) #=> [Money.new(34), Money.new(33), Money.new(33)]
   def split(num)
+    raise ArgumentError, 'need at least one party' if num < 1
+
     calculate_splits(num).sum([]) { |value, count| Array.new(count, value) }
   end
 
-  # Calculate the splits evenly without losing pennies.
-  # Returns the number of high and low splits and the value of the high and low splits.
-  # Where high represents the Money value with the extra penny
-  # and low a Money without the extra penny.
+  # Sum splits allows to perform arithmetic operations on the splits
+  # and get the resulting Money. This method should be used when it's not
+  # a requirement to get the splits. This will skip the Array creation for
+  # better performance.
   #
-  # @param [2] number of parties.
+  # @param [2] number of parties
   #
-  # @return [Hash<Money, Integer>]
+  # @return Money
   #
   # @example
-  #   Money.new(100, "USD").calculate_splits(3) #=> {Money.new(34) => 1, Money.new(33) => 2}
-  def calculate_splits(num)
-    raise ArgumentError, "need at least one party" if num < 1
-    subunits = self.subunits
-    low = Money.from_subunits(subunits / num, currency)
-    high = Money.from_subunits(low.subunits + 1, currency)
+  #   Money.new(100, "USD").sum_splits(3) #=> Money.new(100, "USD")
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -343,29 +343,26 @@ def split(num)
...
-
-    {}.tap do |result|
-      result[high] = num_high if num_high > 0
-      result[low] = num - num_high
+  #   Money.new(100, "USD").sum_splits(3) #=> Money.new(100, "USD")
+  #   Money.new(100, "USD").sum_splits(3) do |split|
+  #     split.allocate([0.9, 0.1]).last
+  #   end #=> Money.new(9.99, "USD")
+  def sum_splits(num)
+    unless block_given?
```
elsom25: Should we have an arg check on `num` here to?

```rb
raise ArgumentError, 'need at least one party' if num < 1
```

lib/money/money.rb
```diff
@@ -343,29 +343,26 @@ def split(num)
...
+      return self
```
elsom25: Not sure on style preferences in this gem, but I do like 1-line guards:

```rb
return unless block_given?
```

lib/money/money.rb
FranOis: It's going to get picked up by `calculate_split` anyway. Not sure how much more we get out of doing the check again.

lib/money/money.rb
FranOis: Being new to Ruby still, I find it harder to read as a single line, but I don't really mind either way.

elsom25: I'd defer to whatever @elfassy and @bdewater prefer

lib/money/money.rb
elsom25: Yeah I always struggle with these kinds of guards, so I'd love extra opinions here; the part that tipped me in favour of adding it at this level is this being a public method with the other being private. 

I think it's nicer when public methods are very specific about what they expect?

lib/money/money.rb
elfassy: we prefer multi-line guards for this gem to make sure the automated code test-coverage calculations work correctly (we have a hard requirement on 100% code coverage)

lib/money/money.rb
```diff
@@ -340,32 +340,33 @@ def allocate_max_amounts(maximums)
...
+  #     split.allocate([0.9, 0.1]).last
+  #   end #=> Money.new(9.99, "USD")
+  def sum_splits(num)
+    raise ArgumentError, 'need at least one party' if num < 1
+
+    unless block_given?
+      return self
     end
+
+    calculate_splits(num).sum { |value, count| yield(value) * count }
```
elfassy: This seems to be a very specific use case. Couldn't people who need this functionality build it themselves?
```ruby
Money.new(100, "USD").sum_splits(3) do |split|
    split.allocate([0.9, 0.1]).last
end
```
vs
```ruby
Money.new(100, "USD").calculate_splits(3).sum do |money, count|
  money.allocate([0.9, 0.1]).last * count
end
```
If we need a shorthand for this, it might be something we add to core instead of this gem

elsom25: I think this errs towards an interface that can be reasonably replicated within `MoneyBag` by narrowing the scope of what's happening?

PR #182, opened 2019-03-01
Forwardable not pre included in newer ruby versions

KaanOzkan:
Noticed this failure in the builds for https://github.com/activemerchant/offsite_payments/pull/313

https://ruby-doc.org/stdlib-2.5.3/libdoc/forwardable/rdoc/Forwardable.html


KaanOzkan:
Can we create a new release?

PR #183, opened 2019-05-27
DO NOT MERGE Tweak Money#<=> to be consistent with Money#==

elsom25:
Right now we have inconsistent responses around equality between `==` (and thus `#eql?`) and `<=>`. This can lead to surprising results.

For example:
```rb
[1] Money.new(2.0) == 2.0
=> false
[2] Money.new(2.0) <=> 2.0
=> 0
```

This PR makes it so that `==`, `eql?`, and `<=>` behave consistently. One side-effect of this however is that `<=>` will return `nil` in more cases (as to be consistent with the old behaviour of `==`), which is why there's a "major" version bump (as major as you can have with pre 1.0).

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -115,16 +115,10 @@ class Money
 
   def <=>(other)
     return unless other.respond_to?(:to_money)
-    arithmetic(other) do |money|
-      value <=> money.value
-    end
-  end
 
-  # TODO: Remove once cross-currency mathematical operations are no longer allowed
-  def ==(other)
-    return false unless other.is_a?(Money)
-    return false unless currency.compatible?(other.currency)
-    super
+    arithmetic(other) do |money|
+      value <=> money.value if currency.compatible?(money.currency)
+    end
   end
   alias_method(:eql?, :==)
 
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -114,10 +114,19 @@ class Money
   end
 
   def <=>(other)
+    return unless other.respond_to?(:to_money)
     arithmetic(other) do |money|
       value <=> money.value
     end
   end
+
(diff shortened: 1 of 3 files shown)


lib/money/money.rb
```diff
@@ -114,10 +114,13 @@ def -@
   end
 
   def <=>(other)
+    return unless other.respond_to?(:to_money)
+
     arithmetic(other) do |money|
-      value <=> money.value
+      value <=> money.value if currency.compatible?(money.currency)
```
elfassy: no need for `currency.compatible?(money.currency)`, this is already handled by `arithmetic(other) do`

elfassy: ah i see you want to make sure it's not equal instead of just having a deprecation. Ignore me

elsom25: AFAICT, `#arithmetic` doesn't guarantee that `money` will be in a matching currency? Likewise, 2 tests fail when removing the condition:

```
rspec ./spec/money_spec.rb:188 # Money does not eql? when currency missmatch
rspec ./spec/money_spec.rb:387 # Money frozen with amount of $1 <=> returns nil when comparing incompatible currency
```

So I think it has to stay for now? Maybe after `#arithmetic` raises on currency mismatch we'll be covered?

lib/money/money.rb
```diff
@@ -114,10 +114,13 @@ def -@
   end
 
   def <=>(other)
+    return unless other.respond_to?(:to_money)
```
elfassy: can we remove this line? I'm worried people are using this for `money == 0` which will now give false even when money is empty

elsom25:
This is unsafe to merge right now without causing significant downstream issues. This kind of change will have to wait until we've formally made currency mismatches deprecated in arithmetic.

PR #185, opened 2019-06-07
Tweak Money#<=> to not raise custom exceptions

elsom25:
Working towards having consistency between `<=>` and `==`, first step is to remove custom exceptions thrown by `<=>`.

### DIFF
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -500,12 +500,12 @@ RSpec.describe "Money" do
     end
 
     describe('any values, non-to_money types') do
-      it { expect { cad_10 <=> nil }.to(raise_error(TypeError)) }
-      it { expect { cad_10 >  nil }.to(raise_error(TypeError)) }
-      it { expect { cad_10 >= nil }.to(raise_error(TypeError)) }
+      it { expect(cad_10 <=> nil).to(eq(nil)) }
+      it { expect { cad_10 >  nil }.to(raise_error(ArgumentError)) }
+      it { expect { cad_10 >= nil }.to(raise_error(ArgumentError)) }
       it { expect(cad_10 == nil).to(eq(false)) }
-      it { expect { cad_10 <= nil }.to(raise_error(TypeError)) }
-      it { expect { cad_10 <  nil }.to(raise_error(TypeError)) }
+      it { expect { cad_10 <= nil }.to(raise_error(ArgumentError)) }
+      it { expect { cad_10 <  nil }.to(raise_error(ArgumentError)) }
     end
   end
 
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -3,7 +3,7 @@ require 'yaml'
 
 RSpec.describe "Money" do
 
-  let (:money) {Money.new(1)}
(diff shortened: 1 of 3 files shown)


spec/money_spec.rb
```diff
@@ -396,6 +388,14 @@
       expect(Money.new(1000, 'USD') <=> Money.new(2000, 'JPY')).to eq(-1)
       expect(Money.new(2000, 'JPY') <=> Money.new(1000, 'USD')).to eq(1)
     end
+
+    it "have the same hash value as $1" do
+      expect(money.hash).to eq(Money.new(1.00).hash)
+    end
+
+    it "does not have the same hash value as $2" do
+      expect(money.hash).to eq(Money.new(1.00).hash)
```
elfassy: test does not match the description

elsom25: This is just moving an existing test; figured that should be fixed in it's own PR?

elsom25: Although its small enough I suppose we can just fold it in

spec/money_spec.rb
```diff
@@ -374,16 +374,8 @@
...
-    end
-
-    it "have the same hash value as $1" do
-      expect(money.hash).to eq(Money.new(1.00).hash)
-    end
-
-    it "does not have the same hash value as $2" do
-      expect(money.hash).to eq(Money.new(1.00).hash)
+    it "<=> returns nil with non-coercible non-money objects" do
+      expect((money <=> nil)).to(eq(nil))
```
elfassy: the current behaviour is
```ruby
[1] pry(main)> Money.new(1) == nil
=> false
[2] pry(main)> Money.new(1) > nil
TypeError: NilClass can't be coerced into Money
from /Users/michaelelfassy/.gem/ruby/2.5.5/gems/shopify-money-0.11.9/lib/money/money.rb:413:in `arithmetic'
```
Are you saying that it makes more sense to have
```
[1] pry(main)> Money.new(1) > nil 
=> nil
```

elsom25: When `<=>` returns nil, if you do `>`, `>=`, `<`, `<=` it'll still raise an exception. BUT `==` returns false

elsom25: And particularly, we want `<=>` to itself never raise an exception (by convention)

elfassy: not sure about the convention, money matches the behaviour of integers:
```
[3] pry(main)> 1 > nil
ArgumentError: comparison of Integer with nil failed
from (pry):3:in `>'
[4] pry(main)> 1 == nil
=> false
```

elsom25: We raise a `TypeError` with a custom error message instead of the standard `ArgumentError`.

That said, actual motivation is related to aspects of certain deprecated behaviour with `MoneyBag` which delegates down to money, and some _specific_ uses of `<=>` directly (vs `==`). This fix moves us _closer_ to `<=>` behaving consistently with `==` (i.e. `<=>` returning `nil` when `==` returns false, instead of `<=>` raising an exception when `==` returns false).

I can add some extra tests before making this change, so we can showcase what the change in behaviour is if that'll help raise confidence?

PR #186, opened 2019-06-08
Add roundrobin_reverse strategy to the money allocation function

itarato:
# The issue

The Money#allocate method is taking care of leftover pennies when dividing the amount to pieces. The current way it's making sure that pennies are not lost is collecting all after rounding up the parts and putting those leftover pennies back to the parts, starting from the first part going clockwise to the end.

This is fine in most cases, however certain scenarios would benefit of controlling this penny fill-up flow.

See the following example (working with integers for sake of simplicity):

- Cart has `2 x $5` product in it, with a `50%` discount -> `2 x $5 - $5`.
- A Buy1Get1Free discount is also added to this line which makes one of the item free

Now remember 5 allocated on 2 is [3, 2], since [floor(2.5), floor(2.5)] will get the leftover 1 penny on the first element.

Since before Buy1Get1Free the line has 2 items totaling $5, the Buy1Get1Free discount will separate them and add 100% on the second part ($2). Once the discount amount is set, the line is divided together with its existing discount: 2 x $5 - $5 will be 1 x $5 - $3 and 1 x $5 - $2 (existing $5 discount is also divided to [3, 2]. And now adding the Buy1Get1Free discount we have 2 lines:
- 1 x $5 - $3 = $2
- 1 x $5 - $2 - $2 = $1

The calculation is correct, however because of the order of penny fills it misses to make the second line free.

# The solution

This is happening because all operation (line splitting, price splitting, discount splitting, discount allocation ...) uses the same strategy for filling up the leftover pennies (left to right). Sometimes for mathematical (and UX) correctness we need to opposite way (reverse fill, from the end) so the pennies "stolen" in one calculation can be given back in another. The example above is a perfect example as 2 different operation (new discount allocation + existing discount split) is acting on exactly the same price in the same flow.

# The change

The PR is introducing a new strategy for penny handling in the `Money#allocate` method - one that fills the pennies from the end towards the beginning (reverse to the current implementation).

There were already references (https://github.com/Shopify/shopify/issues/156354#issuecomment-447496537) of this so I'm inclined to use the namings: `:roundrobin` as current strategy and `:roundrobin_reverse` as the new reverse strategy. The choice on having it a name instead of a boolean flag is that in the future we might introduce another strategy (potentially a `:fair_rounding` strategy which will consider the fragments individually and allocate pennies on greatest fragments first).

The `Money#allocate` by default will act the same as before, no change is required. Using the new strategy is optional - as it will be used only when it's required by the situation.


lib/money/money.rb
```diff
@@ -266,7 +271,11 @@ def fraction(rate)
   #     #=> [#<Money value:20.00 currency:USD>, #<Money value:10.00 currency:USD>]
   #   Money.new(30, "USD").allocate([Rational(2, 3), Rational(1, 3)])
   #     #=> [#<Money value:20.00 currency:USD>, #<Money value:10.00 currency:USD>]
-  def allocate(splits)
+
+  # @example left over pennies distributed reverse order when using roundrobin_reverse strategy
+  #   Money.new(10.01, "USD").allocate([0.5, 0.5], :roundrobin_reverse)
+  #     #=> [#<Money value:5.00 currency:USD>, #<Money value:5.01 currency:USD>]
+  def allocate(splits, strategy = :roundrobin)
```
elfassy: this method is getting somewhat complex. Could you add another commit to this PR to move it to a class, something like:
```ruby
def allocate(splits, strategy = :roundrobin)
  Money::Allocate.new(splits, strategy).perform
end
```

itarato: No problem

itarato: Done https://github.com/Shopify/money/pull/186/commits/27f09024134d84b35fc6d209e6374cf583375388

itarato:
🏓 @elfassy @zacharyreynochiasson

spec/money_spec.rb
```diff
@@ -472,112 +472,16 @@
...
   describe "allocation"do
-    specify "#allocate takes no action when one gets all" do
-      expect(Money.new(5).allocate([1])).to eq([Money.new(5)])
+    specify "#allocate is calculated by Money::Allocator#allocate" do
+      expected = [Money.new(1), [Money.new(1)]]
+      expect_any_instance_of(Money::Allocator).to receive(:allocate).and_return(expected)
+      expect(Money.new(2).allocate([0.5, 0.5])).to eq(expected)
     end
 
-    specify "#allocate does not lose pennies" do
```
elfassy: could you leave one integration test, just to make sure we're not stubbing ourselves into trouble

spec/money_spec.rb
```diff
@@ -472,112 +472,16 @@
   end
 
   describe "allocation"do
-    specify "#allocate takes no action when one gets all" do
-      expect(Money.new(5).allocate([1])).to eq([Money.new(5)])
+    specify "#allocate is calculated by Money::Allocator#allocate" do
+      expected = [Money.new(1), [Money.new(1)]]
+      expect_any_instance_of(Money::Allocator).to receive(:allocate).and_return(expected)
```
elfassy: Let's add `.with` to make sure the arguments received by the allocate method still match
```suggestion
      expect_any_instance_of(Money::Allocator).to receive(:allocate).with([0.5, 0.5], :roundrobin).and_return(expected)
```

spec/money_spec.rb
itarato: No problem, adding.

itarato: Done.

keithbrosnan:
Hi I mentioned this in a linked Issue as Plus merchant having some issues with Automatic discounts. Timeline seems to match with this discussion https://github.com/Shopify/core/issues/6564

elsom25:
@itarato thoughts on opening up a PR to expose strategy choice to `MoneyBag` as well? https://github.com/Shopify/shopify/blob/master/gems/money_bag/lib/money_bag/variadic.rb#L211

(also likely need to bump the money version for the "gem")

itarato:
@elsom25 

> thoughts on opening up a PR to expose strategy choice to MoneyBag as well? 

Yes, that's the next step before the core change. (Version was bumped, thanks for the ping.)

PR #189, opened 2019-09-19
Start testing on Ruby 2.6.x 

XrXr:
Just a few minor changes to get it to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ branches:
   only:
   - master
 rvm:
+  - 2.6
   - 2.5
   - 2.4
   - 2.3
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ branches:
   only:
   - master
 rvm:
+  - 2.6
   - 2.5
   - 2.4
   - 2.3
```
bdewater: We could check in servicesdb if we can kill testing versions we don't rely on internally anymore.

XrXr: I tried checking for it but it's not very easy to do, so I'm going to punt on that :)

PR #192, opened 2019-10-21
Update calculation in amounts_from_splits using Rational and Integer

thomaslacher-shopify:
Update calculation in amounts_from_splits using Rational and Integer to avoid rounding error using BigDecimal.

This fix comes from an issue (https://github.com/Shopify/store/issues/1499) where the allocation of the leftover pennies went to the cheapest items meaning that the calculations felt off.
It caused another problem when trying to refund those product as they were off by a penny (or several if several were refunded).

For example a product with a price of 50$ and a 10% discount could be marked 4.99$ off instead of 5$.

This PR tries to address this by changing the types in the calculation of the max amounts as the rounding was wrong.
For example, in my case I had 5 items ( 2 * 46$ + 3 * 50$) with a discount of 10% off all those items.
The total discount is 24,20$ (or 2420 in subunits).
Calculation for the max amount for an item A at 46$ and item B at 50$ was as follow : 
`frac = (Helpers.value_to_decimal(subunits_to_split * ratio) / allocations).floor`
with 
```
subunits_to_split = 2420.0 {BigDecimal}
ratio = 23/121 (item A) {Rational}
ratio = 25/121 (item B) {Rational}
allocations = 1
subunits_to_split * ratio = 460.0000009 (item A) {BigDecimal}
subunits_to_split * ratio = 499.9999994 (item B) {BigDecimal}
```
Expected answer: 460 / 500

By doing the reverse calculation, it seems like the calcul is working as follow: 
`460.0000009 / 2420.0 = 0.190082645 (item A)`
`499.9999994 / 2420.0 = 0.20661157 (item B)`
When it's supposed to be: 
`ratio = 23/121 = 0.19008264462 (item A)`
`ratio = 23/121 = 0.2066115702 (item B)`

It seems like the ratio is rounded after the 9th decimal which leads to a rounding error afterward.
Case with item A works thanks to `.floor` but case with item B will be off by 1 penny because of `.floor`

I suggest to convert `subunits_to_split` from BigDecimal to Integer since the subunit in all currencies should be an `Integer`. 
That means that the calculation will give a `Rational` instead of a `BigDecimal` avoiding rounding issues.

## 🎩 instruction
Update the Gemfile in core with : `gem "shopify-money", path: "/Users/toto/money_fork"`

Then you can create discount with the same pattern as mentioned above (2 * 46$ items, 3 * 50$ items) and a discount of 10% off those 2 products.
Also make sure that the cheapest items are at the top of the list.

### DIFF
--- a/lib/money/allocator.rb
+++ b/lib/money/allocator.rb
@@ -37,11 +37,8 @@ class Money
     #   Money.new(10.01, "USD").allocate([0.5, 0.5], :roundrobin_reverse)
     #     #=> [#<Money value:5.00 currency:USD>, #<Money value:5.01 currency:USD>]
     def allocate(splits, strategy = :roundrobin)
-      if all_rational?(splits)
-        allocations = splits.inject(0) { |sum, n| sum + n }
-      else
-        allocations = splits.inject(0) { |sum, n| sum + Helpers.value_to_decimal(n) }
-      end
+      splits.map!(&:to_r)
+      allocations = splits.inject(0, :+)
 
       if (allocations - BigDecimal("1")) > Float::EPSILON
         raise ArgumentError, "splits add to more than 100%"
@@ -81,14 +78,11 @@ class Money
       maximums_total = maximums.reduce(Money.new(0, allocation_currency), :+)
 
       splits = maximums.map do |max_amount|
-        next(0) if maximums_total.zero?
+        next(Rational(0)) if maximums_total.zero?
         Money.rational(max_amount, maximums_total)
       end
 
-      total_allocatable = [
-        value * allocation_currency.subunit_to_unit,
-        maximums_total.value * allocation_currency.subunit_to_unit
-      ].min
+      total_allocatable = [maximums_total.subunits, self.subunits].min
 
       subunits_amounts, left_over = amounts_from_splits(1, splits, total_allocatable)
 
@@ -116,10 +110,12 @@ class Money
     end
 
     def amounts_from_splits(allocations, splits, subunits_to_split = subunits)
+      raise ArgumentError, "All splits values must be of type Rational." unless all_rational?(splits)
+
       left_over = subunits_to_split
 
       amounts = splits.collect do |ratio|
-        frac = (Helpers.value_to_decimal(subunits_to_split * ratio) / allocations).floor
+        frac = (subunits_to_split * ratio / allocations.to_r).floor
         left_over -= frac
         frac
       end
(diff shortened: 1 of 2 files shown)


lib/money/allocator.rb
```diff
@@ -119,7 +116,7 @@ def amounts_from_splits(allocations, splits, subunits_to_split = subunits)
       left_over = subunits_to_split
 
       amounts = splits.collect do |ratio|
-        frac = (Helpers.value_to_decimal(subunits_to_split * ratio) / allocations).floor
+        frac = (subunits_to_split * ratio / allocations.to_r).floor
```
itarato: We actually had precision issues using decimal with very low numbers (1..10 divided by 1..10) - and changed to rational for that reason.
In this case I wonder if it would be safer to convert the dividend to rational too, the reason being:

```ruby
refute((BigDecimal(10) / 3.to_r) * 3 === 10)
assert((BigDecimal(10).to_r / 3.to_r) * 3 === 10)
```
And I'm not sure that the dividend can be bigdecimal or not :/

thomaslacher-shopify: From my understanding, `subunits_to_split * ratio` is supposed to always return a Rational since `subunits_to_split` is an Integer and `ratio` is a Rational. 
But I can enforce it just to be sure ! 
@elfassy : What do you think ?

elfassy: that's correct, subunits are always an integer, and integer * rational always gives a rational

spec/allocator_spec.rb
```diff
@@ -127,6 +127,12 @@
         new_allocator(3).allocate_max_amounts([Money.empty, Money.empty, Money.empty]),
       ).to eq([Money.empty, Money.empty, Money.empty])
     end
+
+    specify "#allocate_max_amounts allocates the right amount withou rounding error" do
```
elfassy:
```suggestion
    specify "#allocate_max_amounts allocates the right amount without rounding error" do
```

denislaliberte:
can you recommend  step to  tophat this change ?

lib/money/allocator.rb
itarato: Fair point.

thomaslacher-shopify: @elfassy: Actually I found out that you could send all the parameters in the type that you want (for example when using `allocate`).
I think it would be a good idea to convert the parameters in the type that we want. 
What do you think ?

elfassy: Adding here our IRL conversation:
👍 good catch. Now that you mention it, this method to sum decimals also doesn't make sense
```ruby
if all_rational?(splits)
  allocations = splits.inject(0) { |sum, n| sum + n }
else
  allocations = splits.inject(0) { |sum, n| sum + Helpers.value_to_decimal(n) }
end
```
it can be replaced with with 
```ruby
splits.map!(&:to_r)
allocations = splits.inject(0, :+)
# ...
amounts, left_over = amounts_from_splits(allocations, splits)
```
This private method could raise an argument error is anyone passes a non-rational number. I think this is preferable to converting them in `amounts_from_splits` since there's always a risk in converting a float/decimal back to a rational, and this risk should probably be assumed by a public method.

denislaliberte:
🎩'ed work as expected

PR #199, opened 2020-02-04
Various updates and cleanups

casperisfine:
- Add a shopify-money file so that bundler auto require works
- Add frozen_string_literal: true
- Update test dependencies

### DIFF
--- a/money.gemspec
+++ b/money.gemspec
@@ -1,4 +1,5 @@
 # -*- encoding: utf-8 -*-
+# frozen_string_literal: true
 require_relative "lib/money/version"
 
 Gem::Specification.new do |s|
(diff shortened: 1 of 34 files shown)


money.gemspec
```diff
@@ -14,11 +15,10 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency("bundler", ">= 1.5")
   s.add_development_dependency("simplecov", ">= 0")
-  s.add_development_dependency("rails", "~> 5.0")
+  s.add_development_dependency("rails", "~> 6.0")
```
Edouard-chin: Need to get rid of the 2.3, 2.4  ruby version from the matrix CI

rafaelfranca: And change the `required_ruby_version`

PR #200, opened 2020-04-07
Add Money/MissingCurrency Rubocop rule

bdewater:


lib/rubocop/cop/money/missing_currency.rb
```diff
@@ -0,0 +1,65 @@
...
+          if money_new_without_currency?(node)
+            add_offense(node, message: 'Money is missing currency argument')
+          end
+
+          if to_money_block?(node) || to_money_without_currency?(node)
+            add_offense(node, message: 'to_money is missing currency argument')
+          end
+        end
+
+        def autocorrect(node)
```
elfassy, line 43: do we need autocorrect? Seems a bit dangerous

bdewater: See the readme ;)

PR #203, opened 2020-04-28
Bump Ruby requirement to 2.6 for faster string to BigDecimal path

bdewater:
Some informal testing of `Money.new("7", "CAD")` with benchmark-ips shows roughly a 40% improvement from ~250k to ~350k instructions per second.

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -5,7 +5,6 @@ class Money
   module Helpers
     module_function
 
-    NUMERIC_REGEX = /\A\s*[\+\-]?(\d+|\d*\.\d+)\s*\z/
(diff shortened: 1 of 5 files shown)


lib/money/helpers.rb
```diff
@@ -25,7 +24,11 @@ def value_to_decimal(num)
         when Rational
           BigDecimal(num, MAX_DECIMAL)
         when String
-          string_to_decimal(num)
+          decimal = BigDecimal(num, exception: false)
```
elfassy, line 27: BigDecimal fixed how they handle strings https://github.com/ruby/bigdecimal/issues/63 🎉

PR #204, opened 2020-05-04
Make MissingCurrency cop detect Money.new without any arguments

bdewater:


spec/rubocop/cop/money/missing_currency_spec.rb
```diff
@@ -83,6 +90,11 @@
       expect(new_source).to eq("Money.new(1, 'CAD')")
     end
 
+    it 'corrects Money.new without amount or currency' do
+      new_source = autocorrect_source('Money.new')
+      expect(new_source).to eq("Money.new(0, 'CAD')")
```
elfassy, line 95: Should we go straight to `Money.zero('CAD')`?

elfassy: Actually, probably best not since `Money.zero` without a currency is acceptable (to mean the absence of money) vs `Money.new` where it's not (the specific amount of a currency)

bdewater: Yep. I actually want to make Money.zero less attractive. Soon in another PR!

PR #205, opened 2020-05-05
Remove Money.zero and Money.empty

bdewater:
This is to make very clear that this is a sharp tool, since the warnings mentioned in the documentation have happened. The naming is borrowed from core's multi-currency code. Ideally people would not use this at all and explicitly build Money objects with a currency.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -26,11 +26,6 @@ class Money
     end
     alias_method :from_amount, :new
 
-    def zero(currency = NULL_CURRENCY)
-      new(0, currency)
-    end
-    alias_method :empty, :zero
-
     def parse(*args, **kwargs)
       parser.parse(*args, **kwargs)
     end
--- a/lib/money/null_currency.rb
+++ b/lib/money/null_currency.rb
@@ -1,5 +1,36 @@
 # frozen_string_literal: true
 class Money
+  # A placeholder currency for instances where no actual currency is available,
+  # as defined by ISO4217. You should rarely, if ever, need to use this
+  # directly. It's here mostly for backwards compatibility and for that reason
+  # behaves like a dollar, which is how this gem worked before the introduction
+  # of currency.
+  #
+  # Here follows a list of preferred alternatives over using Money with
+  # NullCurrency:
+  #
+  # For comparisons where you don't know the currency beforehand, you can use
+  # Numeric predicate methods like #positive?/#negative?/#zero?/#nonzero?.
+  # Comparison operators with Numeric (==, !=, <=, =>, <, >) work as well.
+  #
+  # @example
+  #   Money.new(1, 'CAD').positive? #=> true
+  #   Money.new(2, 'CAD') >= 0      #=> true
+  #
+  # Money with NullCurrency has behaviour that may surprise you, such as
+  # database validations or GraphQL enum not allowing the string representation
+  # of NullCurrency. Prefer using Money.new(0, currency) where possible, as
+  # this sidesteps these issues and provides additional currency check
+  # safeties.
+  #
+  # Unlike other currencies, it is allowed to calculate a Money object with
+  # NullCurrency with another currency. The resulting Money object will have
+  # the other currency.
+  #
+  # @example
+  #   Money.new(0, Money::NULL_CURRENCY) + Money.new(5, 'CAD')
+  #   #=> #<Money value:5.00 currency:CAD>
+  #
   class NullCurrency
 
     attr_reader :iso_code, :iso_numeric, :name, :smallest_denomination, :subunit_symbol,
@@ -9,7 +40,7 @@ class Money
       @symbol                = '$'
       @disambiguate_symbol   = nil
       @subunit_symbol        = nil
-      @iso_code              = 'XXX' # Valid ISO4217
+      @iso_code              = 'XXX'
       @iso_numeric           = '999'
       @name                  = 'No Currency'
       @smallest_denomination = 1
--- a/lib/rubocop/cop/money/missing_currency.rb
+++ b/lib/rubocop/cop/money/missing_currency.rb
@@ -20,7 +20,7 @@ module RuboCop
         #
 
         def_node_matcher :money_new, <<~PATTERN
-          (send (const nil? :Money) {:new :from_amount :from_cents} $...)
+          (send (const {nil? cbase} :Money) {:new :from_amount :from_cents} $...)
         PATTERN
 
         def_node_matcher :to_money_without_currency?, <<~PATTERN
@@ -44,9 +44,6 @@ module RuboCop
         end
 
         def autocorrect(node)
-          currency = cop_config['ReplacementCurrency']
-          return unless currency
-
           receiver, method, _ = *node
 
           lambda do |corrector|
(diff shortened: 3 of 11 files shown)


elfassy:
Not sure about this one. Can you explain in your view how `this is a sharp tool`? 

There are two different concepts at play here: `absence of money` vs `amount of money`. In that sense the naming `.zero` doesn't seem ambiguous and already indicates that there's no currency. 

When used in computation with another currency, Money.zero gives way to the currency of the money it is computed with. Since this is consistent with the concept of absence of money, no error or warnings are displayed.
```
Money.new(1, 'USD') + Money.zero #=> Money.new(1, 'USD')
Money.new(1, 'USD') > Money.zero #=> true
```
Moreover `Money.zero` is not something that can't be received as an external input. It needs to be explicitly coded by the developer. In other words `Money.new(amount)` where amount is provided by an external input always has a chance of being 0 or nil or ''. That is a representation of some amount and therefore needs to have a currency. `Money.zero` on the other hand,  is explicitly used when the developer always know this value must be zero.

lib/money/money.rb
```diff
@@ -26,10 +26,15 @@ def new(value = 0, currency = nil)
...
+    # A Money of value zero and no 'real' currency. Warning: this may have
+    # surprising behaviour, such as database validations or GraphQL enum not
+    # allowing the string representation of `NullCurrency`. Prefer
+    # `Money.new(0, currency)` to avoid this ambiguity.
+    #
+    # @return [Money]
+    def ambiguous_currencyless_zero
+      new(0, NULL_CURRENCY)
     end
-    alias_method :empty, :zero
```
elfassy, line 32: won't this break existing code?

bdewater: We are currently on 0.14.1, so we can still discuss and clean up the uglier parts of the API without problems while adhering to SemVer. There's plenty to do before a 1.0 such as making currency a required argument and disallowing cross-currency computation by default, but I felt this was worth tackling as well.

If easy of upgrading is a concern, I'd be happy to write a Rubocop rule for autocorrecting these.

bdewater:
> Can you explain in your view how this is a sharp tool?

I had hoped the accompanying YARD documentation would've explained that 😃 those are real mistakes I've seen happening in production because folks default to `Money.zero` when they didn't need to on code close to an API edge. Since calculation did not happen 'XXX' ends up being the currency code which then fails validation. Taking a bit more effort to have typed something like `Money.new(0, currency)` would have prevented such cases.

> There are two different concepts at play here: `absence of money` vs `amount of money`. 

Not sure I follow this one, can you elaborate? In my view they are both "money", one with clear semantics around handling mixed currencies and one with a special case where it "turns into" the other currency. The amount doesn't matter. 

If the end result of my code ends up being zero in different currencies and I try to sum those (i.e. `Money.new(0, 'USD') + Money.new(0, 'CAD')`) I want it to blow up because I missed some currency conversion step, even if the net result would still be zero. Because the next time, it'll be with positive amounts and then I absolutely want to know.

> In that sense the naming `.zero` doesn't seem ambiguous and already indicates that there's no currency.

Is that as clear for developers who haven't spent a non-trivial amount of time working on this gem as we did? 😉 My perception is more that they are not familiar with the difference and see it just as a quicker way of writing `Money.new(0)`. The fact we still don't have mandatory currency argument plays into this as well.

> `Money.zero` on the other hand, is explicitly used when the developer always know this value must be zero.

I don't understand this one. If it must be zero, that can also be achieved with `Money.new(0, currency)` with the added benefit of currency explicitness? There is no benefit for `Money.zero` that I can see.

I believe the [affordance for errors is too high](https://gmalette.dev/ruby/2020/02/18/affordance-for-errors-pt3.html) with our currently loose handling of currency. `Money.zero` to me is a too easy shortcut and I want to reduce that by giving it a name that properly conveys the tradeoff the developer is making. I really can't imagine situations where I prefer `Money.zero` over `Money.new(0, currency)` if currency is known at the time. For the _rare_ cases you don't this sharp tool is still available, but it should never be the default one to reach for.

elsom25:
>When used in computation with another currency, Money.zero gives way to the currency of the money it is computed with.

I think a name like `Money.ambiguous_currencyless_zero` makes that behaviour very clear? I believe the other _huge_ benefit to this change is the ability to make `Money.zero` **require** a currency. This allows for intent to be expressed much more explicitly for readers of old code?

elfassy:
> I don't understand this one. If it must be zero, that can also be achieved with Money.new(0, currency) with the added benefit of currency explicitness?

Here's an example: say you're accepting an amount/currency from an external source. You don't know what this currency will be but you'd like to verify that the amount is more than zero. Ideally you could do 
```ruby
money = Money.new(params[:amount], params[:currency])
money > Money.zero
```
Of course there are a number of ways to get around this, but all of them are somewhat hacks
```ruby
money > Money.new(0, params[:currency])
money > 0 
money.value > 0
```

>  I've seen happening in production because folks default to Money.zero

🤔 Did they have a legitimate default currency they could use? I guess it does make it tempting to use `Money.zero` instead of a nil value. 

How about `Money.currencyless_zero`  with aliases until we're sure it's no longer used in core? (hehe, compromise :D )

elsom25:
>How about Money.currencyless_zero with aliases until we're sure the old syntax is no longer used in core? (hehe, compromise :D )

What's challenge do you see with `ambiguous` being in the method name (i.e. what makes `.currencyless_zero` preferable to `.amibiguous_currencyless_zero`)? 

Aliases I've got no opinions on, do whatever makes sense ❤️

bdewater:
> Of course there are a number of ways to get around this, but all of them are somewhat hacks
> 
> ```ruby
> money > Money.new(0, params[:currency])
> money > 0 
> money.value > 0
> ```

Interesting! I never considered these hacks. Looking at the code, `Money#coerce` allows any Numeric to be used, and we also explicitly delegate `positive?`/`negative?`/`zero?`/`non_zero?` to the underlying BigDecimal. That seems quite explicit to me?

Comparing Money with Money is also the slowest of the options:
```ruby
require 'benchmark/ips'
require_relative 'lib/money'

MONEY = Money.new("7", "CAD")

Benchmark.ips do |x|
  x.report("> 0") { MONEY > 0 }
  x.report("positive?") { MONEY.positive? }
  x.report("Money.zero") { MONEY > Money.zero }

  x.compare!
end

```

```
                 > 0    87.926k i/100ms
           positive?   121.843k i/100ms
          Money.zero    52.714k i/100ms
                 > 0      1.094M (± 2.8%) i/s -      5.539M in   5.067224s
           positive?      1.592M (± 5.3%) i/s -      8.042M in   5.067871s
          Money.zero    606.064k (± 3.7%) i/s -      3.057M in   5.052252s

Comparison:
           positive?:  1591778.7 i/s
                 > 0:  1094097.4 i/s - 1.45x  slower
          Money.zero:   606064.5 i/s - 2.63x  slower
```

Kinda moot though, a single invocation is fast enough, this only becomes and issue in a loop.

> How about `Money.currencyless_zero` with aliases until we're sure the old syntax is no longer used in core? (hehe, compromise :D )

`currencyless_zero` works for me too with the YARD doc. People read docs right? 😅 It's one of the changes I'd like to introduce, along with enforcing the Rubocop rule for example.

I'm willing to bet we can refactor this in core with Rubocop. If not, I'd prefer the aliases in core and not in the gem. IMO we've had bad defaults in the gem for too long, they carry over into greenfield apps only because we kept accommodating the slower pace of change in the monolith. It creates more work to clean up those new apps after the fact.

elsom25:
@elfassy I'd still really like to understand better what challenges you see with `ambiguous` being in the method name (i.e. what makes `.currencyless_zero` preferable to `.amibiguous_currencyless_zero`)? 

We adopted that name in `MoneyBag` as it conveys what's "risky" about having a currencyless_zero (namely, that you're introducing a level of ambiguity)

elfassy:
@elsom25 IMO there's nothing `ambiguous`, it's currencyless and it's value is zero. If something is ambiguous then we should remove it. If we do we'd need an explicit `Money.new(0, Money::NULL_CURRENCY)` representation.

elsom25:
> EDIT: to be explicit, I'd rather remove Money.zero in favor of Money.new(0, Money::NULL_CURRENCY).

I have mixed feelings. I think there are genuine use-cases for currencyless_Zero, but they're fairly infrequent.

I'd be okay with it going away altogether and only existing in our monolith

bdewater:
> EDIT: to be explicit, I'd rather remove `Money.zero` in favor of `Money.new(0, Money::NULL_CURRENCY)`. Like Bart was saying above there's not much use for this since we can always do the `positive?/negative?/zero?/non_zero?` methods. It also makes it trivial that the currency is wrong in some cases.

🔥 I like that. We'll never really get rid of NullCurrency (think old serialized data) but that's super explicit and autocorrectable with the [MissingCurrency cop](https://github.com/Shopify/money/blob/master/lib/rubocop/cop/money/missing_currency.rb) by changing the default `ReplacementCurrency` setting to `Money::NULL_CURRENCY`.

PR #208, opened 2020-07-28
Update currencies:  kid, mro->mru, vef->ves

cejaekl:
KID is pegged 1:1 to the AUD and circulates interchangeably in Kiribati.
MRU replaces MRO at 10:1.
VES replaces VEF at 100000:1

### DIFF
--- a/config/currency_iso.yml
+++ b/config/currency_iso.yml
@@ -1454,6 +1454,9 @@ mru:
   disambiguate_symbol: MRU
   alternate_symbols: []
   subunit: Khoums
+  # MRU actually has 5 Khoums per Ouguiya, but we treat it as if it has no subunits.
+  # We do this because the Khoum's value is so small that it's not likely to be relevant in an online transaction,
+  # and we suspect that, when our payment partners add support for MRU, they're likely to ignore subunits.
   subunit_to_unit: 1
   symbol_first: false
   html_entity: ''
--- a/config/currency_iso.yml
+++ b/config/currency_iso.yml
@@ -1446,11 +1446,12 @@ mop:
   thousands_separator: ","
   iso_numeric: '446'
   smallest_denomination: 10
-mro:
+mru:
   priority: 100
-  iso_code: MRO
-  name: Mauritanian Ouguiya
+  iso_code: MRU
+  name: Mauritanian New Ouguiya
   symbol: UM
+  disambiguate_symbol: MRU
   alternate_symbols: []
   subunit: Khoums
   subunit_to_unit: 5
@@ -1458,7 +1459,7 @@ mro:
   html_entity: ''
   decimal_mark: "."
   thousands_separator: ","
-  iso_numeric: '478'
+  iso_numeric: '929'
   smallest_denomination: 1
 mur:
   priority: 100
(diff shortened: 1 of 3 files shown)


cejaekl:
Notes (why we're doing this):
  - KID is a prerequisite for permitting signups from Kiribati (https://github.com/Shopify/country_db/issues/401)
  - MRU replaced MRO in December 2017, and is a prerequisite for permitting signups from Mauritania (https://github.com/Shopify/country_db/issues/178)
  - VES replaced VEF in August 2018.  We're not likely to do business in VES anytime soon, but it makes sense to keep our house in order, so we're ready for merchants there when the situation stabilises.

cejaekl:
@bdewater , I have a question that I'm unsure about:

MRU (like MRO before it) subdivides into 5ths.  (It and MGA are the only two currencies in the world that don't use a multiple of ten for their subunits).

I noticed that MRO was entered with `subunit_to_unit: 5`, so I have done the same for MRU.
But https://www.currency-iso.org/en/home/tables/table-a1.html (which doesn't have a good way to handle this case), shows MRU as `minor unit: 2` (which implies `subunit_to_unit: 100`.  Is `subunit_to_unit: 5` going to cause problems interoperating with other platforms?

bdewater:
> MRU (like MRO before it) subdivides into 5ths. (It and MGA are the only two currencies in the world that don't use a multiple of ten for their subunits).

Great question. One of our main use cases here are APIs like Stripe where we're supposed to send amount in the units smallest currency, but [their documentation](https://stripe.com/docs/currencies) doesn't explain if sending `5` (return value of `Money.new(1, "MRO").subunits`) is expected or not.

cejaekl:
Screen Shot 2020-08-18 at 1 05 08 PM

Looking at [some typical prices](https://www.numbeo.com/cost-of-living/country_result.jsp?country=Mauritania&displayCurrency=MRU) in Mauritania, it seems that the subunits (Khoums) are pretty much irrelevant:

Screen Shot 2020-08-18 at 1 10 29 PM

So I guess it's safe to follow Stripe's lead here for MGA, and treat MRU as if it has no subunits.

casperisfine:
Sorry but I have no knowledge nor context on all this. My contribution to this repo are mostly infra structure wise.

config/currency_iso.yml
```diff
@@ -1446,19 +1446,20 @@ mop:
...
-  iso_code: MRO
-  name: Mauritanian Ouguiya
+  iso_code: MRU
+  name: Mauritanian New Ouguiya
   symbol: UM
+  disambiguate_symbol: MRU
   alternate_symbols: []
   subunit: Khoums
-  subunit_to_unit: 5
+  subunit_to_unit: 1
```
cjanke, line 1457: It might be worthwhile to have a comment explaining the "no subunit" thing, like there is below for ves' `smallest_denomination`, in case there are other changes and the git blame doesn't clearly point to your helpful commit message.

config/currency_iso.yml
```diff
@@ -1446,19 +1446,20 @@ mop:
...
-mro:
+mru:
   priority: 100
-  iso_code: MRO
-  name: Mauritanian Ouguiya
+  iso_code: MRU
+  name: Mauritanian New Ouguiya
   symbol: UM
+  disambiguate_symbol: MRU
   alternate_symbols: []
```
plentz, line 1455: Shouldn't we add `MRO` here? (I have no context at all on how  this works and how it's used, just remind of country_db's `alternate_names`)

cejaekl: > Shouldn't we add MRO here?

Ah, but MRO is not an alternate name for MRU.
MRO is the [original Mauritanian Ouguiya](https://en.wikipedia.org/wiki/Mauritanian_ouguiya#First_Ouguiya_(MRO)) (1973-2017).
MRU is the [second (or "new") Mauritanian Ouguiya](https://en.wikipedia.org/wiki/Mauritanian_ouguiya#Second_ouguiya_(MRU)) (2018-present).
1 MRU == 10 MRO
MRO is listed separately in `config/currency_historic.yml`.

Issue #211, opened 2020-09-22
Money#to_s rounds inappropriately


tgwizard:
See the following:

```ruby
[1] pry(main)> raw = BigDecimal("999999999999999999.99")
=> 0.99999999999999999999e18
[2] pry(main)> m = Money.new(raw, 'USD')
=> #<Money value:1000000000000000000.00 currency:USD>
[3] pry(main)> m.to_s
=> "1000000000000000000.00"
[4] pry(main)> m.value
=> 0.99999999999999999999e18
```

The raw value is properly stored as a 18+2 decimal, and internally in the `Money` representation, that's what we use. When we serialize it to a string, we do

https://github.com/Shopify/money/blob/3f51cb84f11e5f3e451806ef4244abafc7a05bed/lib/money/money.rb#L214

This implicitly converts the BigDecimal to a float:

```ruby
[5] pry(main)> raw.to_f
=> 1.0e+18
```

This is a bug in a few ways, but the main issue that happened for me is that the length now differs. Imagine an API that serializes Money values by calling `to_s` on them, and a client that reads these money values and attempts to store them in a database, e.g. using a MySQL `DECIMAL` type. The original money value fits well in a `DECIMAL(20,2)`, but the string-serialized value requires a `DECIMAL(21,2)` to store. The client thus cannot use the same storage column type as the server powering the API, which is confusing.

I'm not sure what the best solution is. If the raw money value has more decimal digits than the currency allows, some kind of rounding or truncation is required. 

@cejaekl @bdewater @Shopify/intl-eng

PR #212, opened 2020-10-08
Handle to_s for amounts larger than float precision

tgwizard:
Fix https://github.com/Shopify/money/issues/211.

The crux of the issue is that some numbers, like `99999999999999.91`, `99999999999999.99`, etc., cannot be represented as floats:

```
irb(main):001:0> 99999999999999.91
=> 99999999999999.9
irb(main):002:0> 99999999999999.99
=> 99999999999999.98
```

The old `to_s` method used the `f` field for `sprintf`, which essentially converts the `Money` amount `BigDecimal` to a float, then to a string.

This PR changes the implementation of `to_s` so that we don't use float as an intermediate representation. Unfortunately, it's a bit messy. It's inspired by https://stackoverflow.com/a/24892562. Feedback and suggestions for improvement more than welcome.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -207,11 +207,20 @@ class Money
   end
 
   def to_s(style = nil)
-    case style
+    units = case style
     when :legacy_dollars
-      sprintf("%.2f", value)
+      2
     when :amount, nil
-      sprintf("%.#{currency.minor_units}f", value)
+      currency.minor_units
+    else
+      raise ArgumentError, "Unexpected style: #{style}"
+    end
+
+    rounded_value = value.round(units)
+    if units == 0
+      sprintf("%d", rounded_value)
+    else
+      sprintf("%d.%0#{units}d", rounded_value.truncate, rounded_value.frac * (10 ** units))
     end
   end
 
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -80,6 +80,26 @@ RSpec.describe "Money" do
     expect(non_fractional_money.to_s(:amount)).to eq("1")
   end
 
+  it "to_s rounds when  more fractions than currency allows" do
+    expect(Money.new("9.999", "USD").to_s).to eq("10.00")
+    expect(Money.new("9.889", "USD").to_s).to eq("9.89")
+  end
+
+  it "to_s does not round when fractions same as currency allows" do
+    expect(Money.new("1.25", "USD").to_s).to eq("1.25")
+    expect(Money.new("9.99", "USD").to_s).to eq("9.99")
+    expect(Money.new("9.999", "BHD").to_s).to eq("9.999")
+  end
+
+  it "to_s does not round if amount is larger than float allows" do
+    expect(Money.new("99999999999999.99", "USD").to_s).to eq("99999999999999.99")
+    expect(Money.new("999999999999999999.99", "USD").to_s).to eq("999999999999999999.99")
+  end
+
+  it "to_s raises ArgumentError on unsupported style" do
+    expect{ money.to_s(:some_weird_style) }.to raise_error(ArgumentError)
+  end
+
   it "as_json as a float with 2 decimal places" do
     expect(money.as_json).to eq("1.00")
   end


tgwizard:
CI seems to fail not because of tests, but because of coverage collection.

lib/money/money.rb
```diff
@@ -207,11 +207,20 @@ def to_d
...
-    case style
+    units = case style
     when :legacy_dollars
-      sprintf("%.2f", value)
+      2
     when :amount, nil
-      sprintf("%.#{currency.minor_units}f", value)
+      currency.minor_units
+    else
+      return nil
```
elfassy, line 216: any reason we would not want to raise here? We expect style to be one of `:legacy_dollars`, `:amount`, or nil.

tgwizard:
@elfassy you're right, `99999999999999.91` as a string/BigDecimal behaves correctly on the master version of Shopify/money, with `to_s` using `%.2f` and all. I used it as an example in the PR description only of something that can't be represented as a float:

```
[3] pry(main)> BigDecimal('99999999999999.91').to_f
=> 99999999999999.9
```

The other example from the PR description, as well as the example for the original issue, behave incorrectly, with rounding errors also with `sprintf`, `%.2f`, and here in Shopify/money.

elfassy:
Looks to me like there's some issue with the precision we're using in the Money::Helper class
```
[35] pry(main)> BigDecimal( 99999999999999.99, 16)
=> 0.9999999999999998e14
[36] pry(main)> Float::DIG
=> 15
[37] pry(main)> BigDecimal( 99999999999999.99, 15)
=> 0.1e15
[38] pry(main)> BigDecimal( 99999999999999.99, 17)
ArgumentError: precision too large.
from (pry):38:in `BigDecimal'
```
For some reason `Float::DIG` is 15, but we're allowed to use a precision of 16. This will at least keep the result consistent between a strong and a float

lib/money/money.rb
```diff
@@ -207,11 +207,20 @@ def to_d
...
+    end
+
+    rounded = round(units)
```
elfassy, line 219: the `value` is already rounded. How about
```ruby
def to_s(style = nil)
  units = (style == :legacy_dollars ? 2 : currency.minor_units)
  if units == 0
    sprintf("%d", value)
  else
    sprintf("%d.%0#{units}d", value.truncate, value.frac * (10 ** units))
  end
end
```

tgwizard: The value isn't always already rounded. Example is `Money.new("1.999", "USD)` (or possibly `Money.new("1.99", "USD) / 2`. These amounts have 3 decimals, and on master they are rendered as `"2.00"` and `"1.00"`, respectively. We need to round for that.

`units = (style == :legacy_dollars ? 2 : currency.minor_units)` is slightly breaking. The old version returns `nil` if `style` is an unexpected value like `:foo`. That's why my code looks like it does. I'm happy to change to what you propose if we're fine with the breaking change.

lib/money/money.rb
elfassy, line 216: >CI seems to fail not because of tests, but because of coverage collection.

I believe this line of code is not tested. Code coverage needs to be 100% on this repo

tgwizard: This line of code existed before (case/switch statement without an `else` clause), but depending on discussion in https://github.com/Shopify/money/pull/212#discussion_r504044321, i'm happy to add a test that asserts an unknown `style` returns `nil`.

lib/money/money.rb
elfassy: sorry i meant the value is rounded when initialized here https://github.com/Shopify/money/blob/master/lib/money/money.rb#L87. You're right though that this won't work for a 3 digit currency we're trying to display using "legacy_dollars" notation. 

I belive you can remove the `to_d` here since these would already all be decimals

tgwizard: Right, I can change `rounded = round(units)` to `rounded_value = @value.round(units)`, and skip the `to_d` parts.

lib/money/money.rb
elfassy: Looks like we were missing a test, probably best to make it raise an argument error if it's not something we expect

spec/money_spec.rb
```diff
@@ -80,6 +80,23 @@
...
+
+  end
+
+  it "to_s does not round when fractions same as currency allows" do
+    expect(Money.new("1.25", "USD").to_s).to eq("1.25")
+    expect(Money.new("9.99", "USD").to_s).to eq("9.99")
+    expect(Money.new("9.999", "BHD").to_s).to eq("9.999")
+  end
+
+  it "to_s does not found if amount is larger than float allows" do
```
bdewater, line 95: found -> round :)

tgwizard:
@elfassy @bdewater I've addressed your comments. Please have another look. (I plan on squashing when merging in the end).

elfassy (approved):

spec/money_spec.rb
```diff
@@ -80,6 +80,26 @@
...
+
+  it "to_s does not round when fractions same as currency allows" do
+    expect(Money.new("1.25", "USD").to_s).to eq("1.25")
+    expect(Money.new("9.99", "USD").to_s).to eq("9.99")
+    expect(Money.new("9.999", "BHD").to_s).to eq("9.999")
+  end
+
+  it "to_s does not round if amount is larger than float allows" do
+    expect(Money.new("99999999999999.99", "USD").to_s).to eq("99999999999999.99")
+    expect(Money.new("999999999999999999.99", "USD").to_s).to eq("999999999999999999.99")
```
elfassy, line 96: please also add `99999999999999.99` decimal use case (will require you to change the precision from 15 to 16)

tgwizard: hmm, you mean 

```ruby
expect(Money.new(BigDecimal( 99999999999999.99, 16), "USD").to_s).to eq("999999999999999999.99")
```

or change `Money::Helpers#value_to_decimal` to use `BigDecimal(num, 16)` instead of `BigDecimal(num, Float::DIG)`?

bdewater (approved):

spec/money_spec.rb
elfassy: just adding, to catch the case where this was changed to 10000...
```ruby
expect(Money.new(99999999999999.99, "USD").to_s).to eq("99999999999999.99")
```

PR #216, opened 2020-11-18
✨ Add subunit override for Stripe

jusleg:
Stripe follows the ISO4217 conventions for most currencies, but not for all, such as ISK. This means that when interacting with the Stripe API and using subunits/from_subunits the amount is off by 2 orders of magnitude.

Since ISO4217 is the only real standard defining this conversion ratio, we keep this as our default behaviour.

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -8,6 +8,10 @@ class Money
     DECIMAL_ZERO = BigDecimal(0).freeze
     MAX_DECIMAL = 21
 
+    STRIPE_SUBUNIT_OVERRIDE = {
+      'ISK' => 100,
+    }.freeze
+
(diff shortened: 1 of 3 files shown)


cejaekl (approved):
This PR is fine by me. 
Personally, my inclination would be to default to doing things Stripe's way; we've done that in the past for [MRU and MGA](https://github.com/Shopify/money/pull/208/commits/3b6ed92496ee66fbf19d9dd5e5188096aa89faa5).  But I would definitely defer to @bdewater on this question.

bdewater (approved):
I've ranted about integers for representing money before in https://github.com/Shopify/money/pull/188#issue-312283502 - using Stripe's (or any other gateway) preference over the ISO standard data isn't a good idea.

Issue #220, opened 2021-01-06
`Money#from_cents` does not take into account non-decimal currencies


dominiquefl:
`Money#from_cents` does not take into account non-decimal currencies prior to creating the money object. The resulting object for non-decimal currencies (such as JPY) will then have a monetary value 1/100 of the actual amount.

https://github.com/Shopify/money/blob/52c160446e0451d410c781f809fa9e1b3523fd6b/lib/money/money.rb#L38-L40

bdewater:
That's correct, this is legacy behaviour from when this gem didn't have currency at all. Use `from_subunits` instead :)

dominiquefl:
Thanks! `from_subunits` is what we need.

PR #221, opened 2021-01-06
Remove legacy cents/from_cents methods

bdewater:
These methods were not doing the right thing and have outlived their usefulness since subunits/from_subunits have existed for a long time - leaving these around just creates confusion as https://github.com/Shopify/money/issues/220 has shown.

I have no idea why we even had a to_liquid method here since we don't do anything Liquid - the couple of apps that use both gems can monkeypatch their required behaviour back in but should not burden dozens of either apps with this concern.

### DIFF
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -247,11 +247,6 @@ RSpec.describe "Money" do
     expect((1.50 * Money.new(1.00))).to eq(Money.new(1.50))
   end
 
-  it "is multipliable by a cents amount" do
-    expect((Money.new(1.00) * 0.50)).to eq(Money.new(0.50))
-    expect((0.50 * Money.new(1.00))).to eq(Money.new(0.50))
-  end
-
   it "is multipliable by a rational" do
     expect((Money.new(3.3) * Rational(1, 12))).to eq(Money.new(0.28))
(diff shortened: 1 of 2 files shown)


spec/money_spec.rb
```diff
@@ -247,11 +247,6 @@
     expect((1.50 * Money.new(1.00))).to eq(Money.new(1.50))
   end
 
-  it "is multipliable by a cents amount" do
-    expect((Money.new(1.00) * 0.50)).to eq(Money.new(0.50))
-    expect((0.50 * Money.new(1.00))).to eq(Money.new(0.50))
-  end
```
fdelache, line 253: This test is still valid (but the name of the test could be better).

fdelache (approved):

spec/money_spec.rb
bdewater: It looks to test the same as the above test? :)

fdelache: Haha, I need to learn how to scroll up 😂

dominiquefl (approved):
looks good; glad you could group it with the `to_liquid` and `cents` method removals. 😄

fredericboivin (approved):
Can't be wrong if the diff is all red

Issue #136, opened 2017-11-14
MoneyAccessor drops currency


ptolts:
I'm wondering if the money accessor should preserve the currency if passed a Money object.

See the additional test failure here: https://github.com/Shopify/money/pull/135

bdewater:
Closing this as https://github.com/Shopify/money/pull/224 has been merged.

PR #224, opened 2021-01-25
Remove MoneyAccessor

bdewater:
This is only being used in one place in all Shopify apps, and it's more clear to see what's going with a regular `attr_accessor` and explicit `Money.new(amount, currency)` usage.

MoneyAccessor is also known not to be safe by dropping currencies, see https://github.com/Shopify/money/issues/136 and https://github.com/Shopify/money/pull/177


fdelache (approved):
🔥

Issue #227, opened 2021-02-19
Duplicate currency entry with the same ISO code 


ttjiaa:
A recent [graphql gem bump PR](https://github.com/Shopify/billing/pull/9275) in Billing which [introduced additional checks to ensured a field isn't defined more than once](https://github.com/rmosolgo/graphql-ruby/commit/48faa8369541f99825d0f68ad63036a5bf6f674f#diff-e218e511a7bd507c9cda3fd9c30e630ba95c89556927fddae92b0e27e55503ffR40-R42) revealed an issue with `Money::Currency.currencies` as there is a pair of currency entries with the same ISO code.

The error is raised on https://github.com/Shopify/billing/blob/dependabot/bundler/graphql-1.12.4/app/graphql/types/currency_code.rb#L8, when the `iso_code` is used to create the fields. There are 2 that are the same with `GHS`, which results in the error `ArgumentError: GHS is already defined for CurrencyCode, please remove one of the definitions.`:
```
[19] pry(main)> (::Money::Currency.currencies.values.map{|x| x["iso_code"]}).count
=> 190
[20] pry(main)> (::Money::Currency.currencies.values.map{|x| x["iso_code"]}).uniq.count
=> 189
```

```
{
   "priority":100,
   "iso_code":"GHS",
   "name":"Ghanaian Cedi",
   "symbol":"₵",
   "disambiguate_symbol":"GH₵",
   "alternate_symbols":[
      "GH₵"
   ],
   "subunit":"Pesewa",
   "subunit_to_unit":100,
   "symbol_first":true,
   "html_entity":"&#x20B5;",
   "decimal_mark":".",
   "thousands_separator":",",
   "iso_numeric":"288",
   "smallest_denomination":1
}
```

```
{
   "priority":100,
   "iso_code":"GHS",
   "name":"Ghanaian Cedi",
   "symbol":"₵",
   "alternate_symbols":[
      "GH¢",
      "GH₵"
   ],
   "subunit":"Pesewa",
   "subunit_to_unit":100,
   "symbol_first":true,
   "html_entity":"&#x20B5;",
   "decimal_mark":".",
   "thousands_separator":",",
   "iso_numeric":"936",
   "smallest_denomination":1
}
```

Based on https://en.wikipedia.org/wiki/ISO_4217#Numeric_codes, it looks like the `"iso_numeric":"288"` one is history while `"iso_numeric":"936"` is current.

[...]

PR #228, opened 2021-03-02
Fix historic Ghanaian Cedi ISO code

acreally-shopify:
The configured historic Ghanaian Cedi ISO code was the same as the
current Ghanaian Cedi ISO code, GHS. However, the currency with numeric
ISO code 288 should have the ISO code GHC, as per:
https://www.currency-iso.org/dam/downloads/lists/list_three.xml

This PR should fix the immediate concern of https://github.com/Shopify/money/issues/227, but there might still be room to explore a solution for ensuring that there aren't duplicate ISO codes in the currency configs.

### DIFF
--- a/config/currency_historic.yml
+++ b/config/currency_historic.yml
@@ -29,7 +29,7 @@ eek:
   smallest_denomination: 5
 ghc:
   priority: 100
-  iso_code: GHS
+  iso_code: GHC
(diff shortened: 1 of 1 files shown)


bdewater:
Thanks! This is good to release now and unblock, we can follow-up with tests later.

PR #231, opened 2021-03-17
Setting to opt-in to v1.0 release

elfassy:
# Why
Make it easy for people to get ready to the v1.0 of money

# What
- raise on deprecation warning
- don't allow not setting a currency (still allows setting one by default)
- to_json works as expected


lib/money/money.rb
```diff
@@ -13,6 +13,16 @@ class Money
   class << self
     attr_accessor :parser, :default_currency
 
+    def opt_in_v1?
+      @opt_in_v1
+    end
+    def opt_in_v1=(value)
+      @opt_in_v1 = value
+      if value
+        Money.active_support_deprecator.behavior = :raise
```
bdewater, line 22: I think we should set this to `:raise` and point people towards either https://blog.testdouble.com/posts/2020-12-15-adding-disallowed-deprecations-to-rails-6-1/ or https://github.com/Shopify/deprecation_toolkit to manage this - no need to reinvent the wheel.

README.md
```diff
@@ -20,6 +20,50 @@ money_column expects a DECIMAL(21,3) database field.
 
     gem 'shopify-money', require: 'money'
 
+## Road to v1.0
```
bdewater, line 23: Maybe an `upgrading.md` instead and only use the readme to show the 'good' behaviour? It can get a little overwhelming otherwise.

lib/money/money.rb
```diff
@@ -76,7 +86,7 @@ def with_currency(new_currency)
 
     def default_settings
       self.parser = MoneyParser
-      self.default_currency = Money::NULL_CURRENCY
+      self.default_currency = opt_in_v1? ? nil : Money::NULL_CURRENCY
```
bdewater, line 89: I've done some digging for https://github.com/Shopify/money/pull/225 and nobody is setting it as far as I can tell - I propose we rip it out altogether.

elfassy: people may use this outside of shopify

lib/money/money.rb
elfassy: Are you saying we should raise by default? 🤔 Not opposed to the idea. I was thinking we'd do this in three steps: 1) optional opt-in 2) opt-in by default, with option to opt-out 3) you're on your own... 😅

bdewater: We already have given the people the ability to optionally opt-in to raising instead of logging over the past couple of years 😉 Some apps have done this but the majority hasn't. Changing the default to `:raise` is in effect the the same as 2) opt-in by default.

lib/money/money.rb
bdewater: They can use the Rubocop rule and get most if not all instances fixed.

lib/money/money.rb
elfassy: I hear you, I think that opt-in option was not well documented so highly possible not many took advantage of it. That being said I agree with you, the time has come for sane defaults. I'll modify this PR to make it opt-out

tomgatzgates (approved):

elfassy:
With the opt-out behaviour in place we can make this the v1.0 release. I think it will make it clear people should read the upgrading info before doing this gem bump. Objections?

spec/money_column_spec.rb
```diff
@@ -166,15 +168,15 @@ class MoneyClassInheritance2 < MoneyWithCustomAccessors
     let(:amount) { 'foo' }
 
     it 'raises a deprecation warning' do
-      expect { subject }.to raise_error(ActiveSupport::DeprecationException)
+      expect { subject }.to raise_error(ArgumentError)
```
elfassy, line 171: fix test description

lib/money/money_parser.rb
```diff
@@ -82,9 +82,12 @@ def extract_amount_from_string(input, currency, strict)
     number = number.to_s.strip
 
     if number.empty?
-      raise MoneyFormatError, "invalid money string: #{input}" if strict
-      Money.deprecate("invalid money strings will raise in the next major release \"#{input}\"")
-      return '0'
+      if strict || !Money.config.legacy_support?
```
elfassy, line 85: inverse logic

bdewater:
> With the opt-out behaviour in place we can make this the v1.0 release. I think it will make it clear people should read the upgrading info before doing this gem bump. Objections?

I don't see the need for haste here to commit to 1.0 just yet. I'd like to backport core's ActiveJob adapter at least first and there's probably 2-3 other things I'd love to clean up :)

bdewater:
Among other things, I question the assumption made in this PR we need any global configuration at all. I will extend https://github.com/Shopify/money/pull/225 with a few commits as an alternative proposal.

elfassy:
@bdewater 
> I don't see the need for haste here to commit to 1.0 just yet
I agree. A pre-release v1.0-pre would allow us to add what's missing for the full 1.0 release. I'd really want to avoid staying <1.0 if we're updating the defaults.

> I question the assumption made in this PR we need any global configuration at all
I don't think it's possible to have the new `to_json` behaviour without a configuration setting to enable it. This was the main driver for this PR and something the billing team requires. I would really like to avoid having to maintain both >1.0 and <1.0 releases. New apps that use this gem should start right away with the correct defaults

> I will extend #225 with a few commits as an alternative proposal.
I don't think removing the `default_currency` option is the way to go. Many apps use a single currency and it would be impractical for them to add it each time. A default currency is a basic feature of any money gem. By the number of stars/followers I'd assume a couple use this feature. The default NullCurrency is what we want to remove IMO

tomgatzgates:
>core's ActiveJob adapter 

Whats this referring to OOI?

elfassy:
@tomgatzgates I think something like this https://github.com/Shopify/banking/pull/3940/files but as part of this gem

tomgatzgates:
Ah gotcha, we spiked the same https://github.com/Shopify/billing/pull/9522/files#diff-baec507276cecd998b6669e1167cd8eb4dae2c56501334741f7e396a7cab3dfa

bdewater:
> I don't think it's possible to have the new to_json behaviour without a configuration setting to enable it. This was the main driver for this PR and something the billing team requires. 

Having reviewed https://github.com/Shopify/billing/pull/9522 now - I see a disconnect between the problem being solved there and this pull request. Yes, we should take care when changing the serialization formats for running systems in a compatible manner, a global configuration might be a good way to go here. 

But as it stands this PR does so much more: a set of new defaults in the form of `legacy_support` (that could be opted-out of), doing a similar thing as the existing Active Support deprecator but with less fine-grained control. It makes discussing the merits of the various changes not easy.

The way I read the suggestion in that PR to upstream the monkey patch, I am imagining something more like:
```diff
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -11,7 +11,7 @@ class Money
   def_delegators :@value, :zero?, :nonzero?, :positive?, :negative?, :to_i, :to_f, :hash
 
   class << self
-    attr_accessor :parser, :default_currency
+    attr_accessor :parser, :default_currency, :json_format
 
     def new(value = 0, currency = nil)
       value = Helpers.value_to_decimal(value)
@@ -77,6 +77,7 @@ class Money
     def default_settings
       self.parser = MoneyParser
       self.default_currency = Money::NULL_CURRENCY
+      self.json_format = nil
     end
   end
   default_settings
@@ -230,11 +231,16 @@ class Money
   end
 
   def to_json(options = {})
-    to_s
+    as_json.to_json
   end
 
-  def as_json(*args)
-    to_s
+  def as_json(options = nil)
+    format = options&.dig(:format) || Money.json_format
+    if format == :legacy
+      to_s
+    else
+      { value: to_s, currency: currency.iso_code }
+    end
   end
```

I have strong opinions on these global config values, but that doesn't need to hold this up some JSON changes. Can we scope down this PR to focus on just that use case for now and hash (pun intended) out the path to 1.0 in follow-up PRs?

elfassy:
Spoke with @bdewater IRL. Here's what we agreed upon:
- best practice by default
- going with 1.0.pre since we're changing the defaults (breaking change). A couple more things will go in a 1.0 release
- global configuration with per-things: legacy_json_format/legacy_default_currency/legacy_default_parser etc. signal clearly the intent to get away from these settings

lib/money/money.rb
```diff
@@ -230,11 +240,19 @@ def to_s(style = nil)
...
+    if Money.config.legacy_json_format
+      to_s
+    else
+      as_json.to_json
+    end
   end
 
-  def as_json(*args)
-    to_s
+  def as_json(*_args)
```
bdewater, line 250: What do you think about [allowing this to be set per call site](https://github.com/Shopify/money/pull/231#issuecomment-802468335), falling back to the global default if that's not done? This allows for more granular transition.

UPGRADING.md
```diff
@@ -0,0 +1,57 @@
...
+```ruby
+Money.configure do |config|
+  config.legacy_default_currency!
+  config.legacy_deprecations!
+  config.legacy_json_format!
+  #...
+end
+```
+
+Remove each legacy setting making sure your app functions as expected.
```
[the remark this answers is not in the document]
elfassy: please do 👍

lib/money/helpers.rb
```diff
@@ -46,16 +46,20 @@ def value_to_currency(currency)
         currency
       when nil, ''
         default = Money.current_currency || Money.default_currency
-        raise(ArgumentError, 'missing currency') if default.nil? || default == ''
+        raise(Money::Currency::UnknownCurrency, 'missing currency') if default.nil? || default == ''
```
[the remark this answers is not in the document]
elfassy: This specific exception is handled by `legacy_default_currency` which will set the default currency to a NullCurrency (if none is set) and not raise here. 
I'm ok with changing this to instead have a deprecation here when `legacy_deprecation` is enabled, otherwise raise.
Not sure which is easier to manage, I'm good either way

lib/money/money.rb
```diff
@@ -230,11 +240,19 @@ def to_s(style = nil)
   end
 
   def to_json(options = {})
-    to_s
+    if Money.config.legacy_json_format
+      to_s
+    else
+      as_json.to_json
```
elfassy, line 246:
```suggestion
      as_json.to_json(options)
```

lib/money/money.rb
elfassy: `as_json` already accepts [many options](https://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html#method-i-as_json) `root`, `include`, `except`, etc. We'd need to make our option as part of the hash in order to not break the signature. If we're comfortable with potential key conflicts, then we can go ahead

spec/spec_helper.rb
```diff
@@ -17,7 +17,7 @@
 require 'money'
 
 Money.active_support_deprecator.behavior = :raise
-Money.default_currency = Money::Currency.new('CAD')
+Money.config.default_currency = Money::Currency.new('CAD')
```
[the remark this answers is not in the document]
elfassy: 💯 agree with not using a default currency in the tests, it definitely makes them harder to reason about 👍 
Not yet convinced we should get rid of the default currency all together though, there are definitely some apps that only cater to one currency. Let's continue the discussion on that PR 👍

spec/spec_helper.rb
```diff
@@ -68,3 +68,18 @@ def missing_methods
     expected.instance_methods - actual.instance_methods
   end
 end
+
+
+def configure(default_currency: nil, legacy_json_format: nil, legacy_deprecations: nil, legacy_default_currency: nil, parser: nil)
+  old_config = Money.config.dup
```
elfassy, line 74: don't think I need the dup here
```suggestion
  old_config = Money.config
```

tomgatzgates:
Anything I can do to help bump this?

elfassy:
@tomgatzgates thanks for the offer to help. I rebase and addressed comments, should be good to ship 
cc @bdewater

lib/money/money_parser.rb
```diff
@@ -82,9 +82,12 @@ def extract_amount_from_string(input, currency, strict)
     number = number.to_s.strip
 
     if number.empty?
-      raise MoneyFormatError, "invalid money string: #{input}" if strict
-      Money.deprecate("invalid money strings will raise in the next major release \"#{input}\"")
-      return '0'
+      if Money.config.legacy_deprecations && !strict
+        Money.deprecate("invalid money strings will raise in the next major release \"#{input}\"")
+        '0'
```
christhomson, line 87: Does this still need to be `return '0'`?

UPGRADING.md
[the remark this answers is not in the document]
bdewater: Haven't been able to find time for this, nor will I be able to in the next couple of weeks. 

Roughly speaking, what I had in mind as an upgrade guide if someone wants to take it on:
1. use [the included Rubocop cops](https://github.com/Shopify/money#rubocop) to autocorrect what can be found with static analysis
2. record deprecations using https://github.com/Shopify/deprecation_toolkit/ for anything not covered by 1 to prevent new deprecations to be included.

lib/money/money_parser.rb
christhomson: Sorry - I meant, this line is just `'0'` - should it be `return '0'`? Otherwise this '0' is just immediately thrown out?

elfassy: oh sorry! Will open a new PR to fix this

PR #237, opened 2021-10-14
Parser refactoring

bdewater:
This PR does a few things based on a couple of use cases that emerged in Shopify:

First it introduces a parser that is locale-aware and can properly parse amounts with 3 digits, for the currencies Bahraini dinar, Iraqi dinar, Jordanian dinar, Kuwaiti dinar, Libyan dinar, Omani rial and Tunisian dinar. This parser also normalizes fullwidth characters, this is useful for e.g. Japanese users where their phone keyboards might lead them to entering these characters. In the future normalizing other digits from the [Unicode Character Category 'Number, Decimal Digit'](https://www.compart.com/en/unicode/category/Nd) could be possible but for now we're just solving the use case we've actually run into.

Second, it introduces a 'simple' parser for API cases (which can also handle 3 decimal currencies) in order to be used in conjunction Shopify's gem that adds type coercion to controller parameters. We had a recent issue where the Fuzzy parser was used in a scenario like this, multiplying amounts by a 1000 , and requiring an extensive data cleanup effort.

```ruby
Money::Parser::Fuzzy.parse("123.456", "USD")
=> #<Money value:123456.00 currency:USD>

Money::Parser::Simple.parse("123.456", "USD")
=> #<Money value:123.46 currency:USD>

### CODE REVIEW

File: lib/money/parser/fuzzy.rb [hunk 1]
```diff
@@ -0,0 +1,174 @@
...
+
+        if last_digits_decimals?(digits, marks, currency)
+          "#{digits.first}.#{digits.last}"
+        else
+          "#{digits.first}#{digits.last}"
+        end
+      end
+
+      def last_digits_decimals?(digits, marks, currency)
+        # Thousands marks are always different from decimal marks
```
cejaekl, line 144: Nitpick:  `Grouping` marks, not `Thousands` marks.  In locales that adopt the "Chinese" 4-digit grouping system, one group is a myriad, not a thousand.

lib/money/parser/locale_aware.rb
```diff
@@ -0,0 +1,52 @@
...
+        # @param decimal_separator [String]
+        # @return [Money, nil]
+        def parse(input, currency, strict: false, decimal_separator: decimal_separator_proc.call)
+          raise ArgumentError, "decimal separator cannot be nil" unless decimal_separator
+
+          currency = Money::Helpers.value_to_currency(currency)
+          return unless currency
+
+          normalized_input = input
+            .tr('－０-９．，', '-0-9.,')
```
cejaekl, line 39: Should we also convert `、` (U+3001) and `､` (U+FF64) to `,` (U+002C)?

lib/money/parser/fuzzy.rb
bdewater: TIL!

elfassy:
What do you think of moving the parsing functionality to a new gem? A bit like rubyMoney which split its parsing to https://github.com/RubyMoney/monetize

lib/money/parser/simple.rb
```diff
@@ -0,0 +1,28 @@
...
+        # locale into consideration.
+        # @param input [String]
+        # @param currency [String, Money::Currency]
+        # @param strict [Boolean]
+        # @return [Money, nil]
+        def parse(input, currency, strict: false)
+          currency = Money::Helpers.value_to_currency(currency)
+          return unless currency
+
+          amount = BigDecimal(input, exception: false)
```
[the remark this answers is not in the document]
bdewater: Calling Money directly adds the complexity of an app potentially having opted into deprecated behaviour and returning `0` instead of `nil` for invalid input, this doesn't work with the expectations from our parameter type coercion gem that is used in Shopify core and other apps.

This parser is meant for well-formatted data so I am wary of normalizing here to paper over bugs from other people. We can loosen restrictions later if we learn it's needed (the opposite is much harder).

Good call on the scientific notation, I added something for that 😄

lib/money/parser/locale_aware.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+class Money
+  module Parser
+    class LocaleAware
+      @decimal_separator_proc = ->() { ::I18n.translate("number.currency.format.separator") }
```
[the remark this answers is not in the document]
bdewater: I put this in a Railtie instead. If we're going to clean this up, which I think is a good idea but outside of the scope for this PR, I don't see what advantage a separate gem gives us at Shopify. I do see disadvantages like extra work for coordinating releases.

lib/money/money.rb
```diff
@@ -33,10 +33,6 @@ def new(value = 0, currency = nil)
     end
     alias_method :from_amount, :new
 
-    def parse(*args, **kwargs)
```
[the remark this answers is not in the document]
elfassy: or add the version change to this PR

lib/money/parser/simple.rb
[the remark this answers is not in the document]
elfassy: > opted into deprecated behaviour and returning 0 instead of nil

If we're already expecting a well formatted string, could we get them to not opt-in to that deprecated behaviour, instead of adding a new parser? IMO nil -> 0 should already never be happening and is fairly easy to do manually if needed `Money.new(amount || 0)`

lib/money/parser/simple.rb
```diff
@@ -15,8 +17,8 @@ def parse(input, currency, strict: false)
           currency = Money::Helpers.value_to_currency(currency)
           return unless currency
 
-          amount = BigDecimal(input, exception: false)
-          if amount
+          coerced = input.to_s
+          if SIGNED_DECIMAL_MATCHER.match?(coerced) && (amount = BigDecimal(coerced, exception: false))
             Money.new(amount, currency)
```
elfassy: nit: we should avoid using an assignment on the second part of the boolean

bdewater: Our style guide [allows it](https://ruby-style-guide.shopify.dev/#:~:text=wrap%20assignment%20in%20parentheses%20when%20using%20its%20return%20value%20in%20a%20conditional%20statement.). I'll look into adding our Rubocop rules to this repo because linters can do this better than humans 😄

bdewater: This will be a fair bit of work: https://github.com/Shopify/money/pull/240 will circle back after this and some other outstanding PRs are merged to minimize disruption.

lib/money/parser/simple.rb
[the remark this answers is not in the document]
bdewater: I am not sure I follow. Are you suggesting to remove the deprecation for invalid input for Money.new? 

Assuming an app hasn't opted in to the deprecated path, Money.new accepts more input than we want for APIs. Money.new will raise on invalid input where our typed parameter code (see related PR) and GraphQL types need a `nil` returned. We could rescue for those cases but that's less performant for most inputs.

I don't understand how `Money.new(amount || 0)` is of use, but if you meant `Money::Parser::Simple.parse(amount) || 0` then yes that's another reason to prefer returning `nil`.

kogulko:
Left comment about additional test
LGTM

PR #241, opened 2021-12-07
Introduce Money#to_formatted_s for consistency with Active Support

casperisfine:
Ref: https://github.com/rails/rails/pull/43772

Most if not all `to_s(format)` are deprecated in Rails 7, having the same `to_formatted_s` interface makes it easier for polymorphic callsites.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -222,14 +222,14 @@ class Money
     value
   end
 
-  def to_s(style = nil)
+  def to_formatted_s(style = nil)
     units = case style
     when :legacy_dollars
       2
     when :amount, nil
       currency.minor_units
     else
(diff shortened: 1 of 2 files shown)


etiennebarrie (commented):
Should we also deprecate it and remove it down the line? To get the optimization?
> polymorphic callsites

Are there callsites where you don't know which object we're using and calling `to_s` with a format?

casperisfine:
> Should we also deprecate it and remove it down the line? To get the optimization?

No need, having `to_s` on arbitrary classes is fine. It's only if you define it on one of the optimized classes (`Numeric`, `Module`, `Symbol`, etc) that it disables the optimization.

I'm fine with deprecating though, but I figured it wasn't strictly required so limited my change to the smallest possible.

PR #246, opened 2022-07-11
Update the database field for money_column

AlexMooney:
The gh_pages index.md has a different column specification from the main README.md.


AlexMooney:
I've signed the CLA, but don't seem to have permissions to rerun that check.

PR #256, opened 2023-01-24
Allow splat arg in `Money/MissingCurrency` cop

sambostock:
This stops the `Money/MissingCurrency` cop from considering the following an offense:

```diff
 value_and_currency = [1, 'CAD']
 Money.new(*value_and_currency)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Money is missing currency argument
```

It is unlikely that someone would use a splat without knowing it contains both an amount and currency, so we should allow it.

### DIFF
--- a/lib/rubocop/cop/money/missing_currency.rb
+++ b/lib/rubocop/cop/money/missing_currency.rb
@@ -32,7 +32,8 @@ module RuboCop
         PATTERN
 
         def on_send(node)
-          money_new(node) do |_amount, currency_arg|
+          money_new(node) do |amount, currency_arg|
+            return if amount&.splat_type?
             return if currency_arg
 
             add_offense(node, message: 'Money is missing currency argument')
--- a/spec/rubocop/cop/money/missing_currency_spec.rb
+++ b/spec/rubocop/cop/money/missing_currency_spec.rb
@@ -26,6 +26,13 @@ RSpec.describe RuboCop::Cop::Money::MissingCurrency do
       RUBY
     end
 
+    it 'does not register an offense for Money.new with splat argument' do
+      expect_no_offenses(<<~RUBY)
+        value_and_currency = [1, 'CAD']
+        Money.new(*value_and_currency)
+      RUBY
+    end
+
     it 'registers an offense and corrects for Money.new without a currency argument' do
       expect_offense(<<~RUBY)
         Money.new


lib/rubocop/cop/money/missing_currency.rb
```diff
@@ -32,7 +32,8 @@ class MissingCurrency < Cop
         PATTERN
 
         def on_send(node)
-          money_new(node) do |_amount, currency_arg|
+          money_new(node) do |amount, currency_arg|
+            return if amount&.splat_type?
```
davidpatters0n, line 36: `splat_type?` is this a method within rubocop AST?

sambostock: Yup! Every `Node` type responds to all the `X_type?` methods.

Using `ruby-parse -e "..code..."`, we can inspect the AST:

<table><tr><th>Code</th><th>AST</th><th><code>amount</code></th><th><code>currency_arg</code></th></tr><tr><td>

```ruby
Money.new
```

</td><td>

```ruby
(send
  (const nil :Money) :new)
```

</td><td>

```ruby
nil
```

</td><td>

```ruby
nil
```

</td></tr><tr><td>

```ruby
Money.new(1)
```

</td><td>

```ruby               
(send
  (const nil :Money) :new
  (int 1))
```

</td><td>

```ruby
(int 1)
```

</td><td>

```ruby
nil
```

</td></tr><tr><td>

```ruby
Money.new(1, "CAD")
```

</td><td>

```ruby
(send
  (const nil :Money) :new
  (int 1)
  (str "CAD"))
```

</td><td>

```ruby
(int 1)
```

</td><td>

```ruby
(str "CAD")
```

</td></tr><tr><td>

```ruby
amount_and_currency = [1, "CAD"]
Money.new(*amount_and_currency)
```

</td><td>

```ruby
(begin
  (lvasgn :amount_and_currency
    (array
      (int 1)
      (str "CAD")))
  (send
    (const nil :Money) :new
    (splat
      (lvar :amount_and_currency))))
```

</td><td>

```ruby
(splat
  (lvar :amount_and_currency))
```

</td><td>

```ruby
nil
```

</td></tr><tr><td>

</tr></table>

From there, we can see the only cases we care about flagging in this Cop are when
- `currency_arg` is `nil`, and
- `amount` is not a `splat` type node

elfassy (approved):

Issue #250, opened 2022-11-16
to_money doesn't handle 3 decimal places all the time


mark-shan:
In `shopify/billing`, we were running a task to buld money objects from a csv. Mode was returning money objects with 3 decimal places which we assumed was harmless. 
We expected $29 of credits to be created, but $29000 ended up being created.

The PR where it was detected https://github.com/Shopify/billing/pull/15628
The line if code which converts csv value to money https://github.com/Shopify/billing/blob/a89cebbecd837c13295fa73554e81a73bf3ee843/app/jobs/maintenance/partially_reverse_or_refund_payables_task.rb#L20

<img width="652" alt="Screenshot 2022-11-16 at 17 40 43" src="https://user-images.githubusercontent.com/35412876/202313297-69524b4a-3a5a-4736-ba82-ecc7538498c6.png">

mark-shan:
> `to_money` is supposed to be banned, I thought, for this reason:
> 
> ```
> [1] pry(main)> "29.000".to_money("USD")
> => #<Money value:29000.00 currency:USD>
> [2] pry(main)> Money.new("29.000", "USD")
> => #<Money value:29.00 currency:USD>
> ```
> 
> Do we know how this got through?

cautiontape bot was giving the opposite suggestion for the original PR for this maintenance task https://github.com/Shopify/billing/pull/11435#issuecomment-953130003

@AndreiUngur did recently suggest we remove that 😄

elfassy:
The`Money::Parser::Fuzzy` parser is deprecated but for legacy reasons is still being used by `.to_money(...)`. Please replace all `to_money` with `Money::Parser::Simple.parse(value, currency)`

elfassy:
This seems serious enough that I'm good with introducing a breaking change. Scanning through the code, I don't see many people using `.to_money` with data coming directly from the user (which would maybe benefit from the fuzzy matching). I'll open a PR

PR #264, opened 2023-06-20
Add a Money/UnsafeToMoney cop

lavoiesl:
Prevents a potentially dangerous use of the Fuzzy parser.

See comment for detail and https://github.com/Shopify/money/issues/250#issuecomment-1345452314 for context


AndreiUngur (approved):
LGTM 👍 

Seems like there are no "code-owners" defined? But I think this change should be harmless given that it's adding an optional cop. Should also be safe to revert if anything;

I wonder if we should link this comment in the code as a justification for the cop? I really don't feel strongly : https://github.com/Shopify/money/issues/250#issuecomment-1345452314

cc @elfassy

elfassy (approved):

bslobodin (commented):
I would also suggest including the new cop in the README,

https://github.com/Shopify/money#rubocop

elfassy:
PR to deprecated the legacy behaviour for new apps https://github.com/Shopify/money/pull/267

PR #267, opened 2023-06-20
update `to_money` to have the same behaviour as `Money.new`

elfassy:
# Why

To avoid gotchas, `string.to_money("USD")` should behave exactly the same as `Money.new(string, "USD")`
Also the deprecation warning will help us search the logs to know how often and who is using the method incorrectly

# What
Make the change but keep the old behaviour under the `legacy_deprecations` option... until we're confident we can remove it. For those with that legacy support, they can rely on the newly added rubocop rule to help them migrate away from the old `.to_money`

### DIFF
--- a/lib/money/core_extensions.rb
+++ b/lib/money/core_extensions.rb
@@ -14,6 +14,11 @@ end
 #   '100.37'.to_money => #<Money @cents=10037>
 class String
   def to_money(currency = nil)
-    Money::Parser::Fuzzy.parse(self, currency)
+    if Money.config.legacy_deprecations
+      Money.deprecate("`#{self}.to_money` will raise an ArgumentError in the next major release")
+      Money::Parser::Fuzzy.parse(self, currency)
+    else
+      Money.new(self, currency)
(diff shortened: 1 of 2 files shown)


lib/money/core_extensions.rb
```diff
@@ -14,6 +14,11 @@ def to_money(currency = nil)
 #   '100.37'.to_money => #<Money @cents=10037>
 class String
   def to_money(currency = nil)
-    Money::Parser::Fuzzy.parse(self, currency)
+    if Money.config.legacy_deprecations
+      Money.deprecate("`#{self}.to_money` will raise an ArgumentError in the next major release")
+      Money::Parser::Fuzzy.parse(self, currency)
+    else
+      Money.new(self, currency)
+    end
```
csalvato, line 22: This means that the gem consumer will need to set legacy_deprecations to true for this to maintain the old behavior, correct?

csalvato (approved):

lavoiesl (approved):

PR #268, opened 2023-06-21
Fix UnsafeToMoney for receiver-less calls

pjambet:
The cop used to match on "receiver-less" calls, and crash when trying to use the autocorrect logic.

The error was:

```
RuboCop::ErrorWithAnalyzedFileLocation:
  cause: #<NoMethodError: undefined method `source' for nil:NilClass>
```

Coming from [this line][1]

---

I decided to not flag these calls, in my example it was a private method called `to_money`, and was therefore not an offense that the cop tries to catch.

---

[1]:https://github.com/Shopify/money/blob/b35448a60c65da08df73f700447b76d3360027b4/lib/rubocop/cop/money/unsafe_to_money.rb#L26

### DIFF
--- a/lib/rubocop/cop/money/unsafe_to_money.rb
+++ b/lib/rubocop/cop/money/unsafe_to_money.rb
@@ -16,7 +16,7 @@ module RuboCop
 
         def on_send(node)
           return unless node.method?(:to_money)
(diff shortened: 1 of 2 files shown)


lavoiesl (approved):
Good catch 👍🏼

PR #278, opened 2024-01-05
Better error if `Money#allocate` called with an empty array

ghiculescu:
At the moment if you do `Money.new(1, "USD").allocate([])` it will raise with:

```
lib/money/allocator.rb:138:in `%': divided by 0 (ZeroDivisionError)
```

Since this is an invalid input, it's better to identify that and raise earlier with a more clear message.


ghiculescu:
Done @elfassy

PR #279, opened 2024-01-26
Drop support for initialization with another money object.

csalvato:
We have been using `Shopify/money` on `Shopify/billing for a while.  For much of that time, we have been avoiding instatiating new Money objects because it is error prone. With that in mind, we have used `to_money` with preference.

Recently, `to_money` has also been found to be dangerous and error prone.

With this PR, we aim to make it so that:
1. Any consumer of this gem can totally avoid using the unsafe `to_money` operation
2. Instantiating new objects will always be safe.

## Why is `Money.new` also unsafe?

`Money.new` has no safeguards in place to ensure that, when providing a `Money` object to the initializer, the currencies match.

That creates a vector for easy-to-produce but hard-to-debug errors. For example:

```rb
money_obj = Money.new(5, 'EUR')
Money.new(money_obj, 'USD')`=> #<Money value:5.00 currency:USD>
```

In this case the new money object will be $5 USD instead of being converted.

The simplest way to avoid these issues is to not allow Money objects to be passed into `Money.new`:

```rb
money_obj = Money.new(5, 'EUR')
Money.new(money_obj, 'USD')` # => Raises `ArgumentError`
Money.new(money_obj.value, 'USD')` # => #<Money value:5.00 currency:USD>
```

We believe this makes sense because:
1. It creates a better-defined interface for the initializer. It will only take scalars - it won't take objects.
2. It solves the problem without adding a ton of responsibilities to the gem, like handing currency conversion rates
3. When used, the initialization of a new money object from an old one is much more explicit. It's clear that the *value* is being used, but the currency is being ignored.
4. It makes migrating from the old version to the new version dead simple - whenever you pass in a money object, just call `.value` on it to get the value.


lib/money/money.rb
```diff
@@ -137,13 +137,8 @@ def -(other)
...
 
   def *(numeric)
-    unless numeric.is_a?(Numeric)
-      if Money.config.legacy_deprecations
-        Money.deprecate("Multiplying Money with #{numeric.class.name} is deprecated and will be removed in the next major release.")
-      else
-        raise ArgumentError, "Money objects can only be multiplied by a Numeric"
-      end
-    end
+    raise ArgumentError, "Money objects can only be multiplied by a Numeric" unless numeric.is_a?(Numeric)
```
csalvato, line 140: The money multiplication feature was deprecated years ago, so I removed it to avoid having to maintain test coverage for it in this PR with this change.

elfassy:
This change makes sense to me. Can you run core's test suite with this branch to see if this creates unforeseen issues 🙏

lib/money/money.rb
elfassy: confirmed, we're not seeing this in core
https://app.bugsnag.com/shopify/shopify-production/errors?filters[event.since]=all&filters[error.status]=open&filters[event.class]=ActiveSupport%3A%3ADeprecationException&filters[event.message]=Multiplying%20Money

csalvato:
> This change makes sense to me. Can you run core's test suite with this branch to see if this creates unforeseen issues 🙏

Will do, and will report back

csalvato:
It's running now. Posting here for an easy link to see the CI when I come back to check on this PR.

https://github.com/Shopify/shopify/pull/478715

csalvato:
I am not sure on the process here, but Core is not on the current version of this gem.

I believe the process is to upgrade the gem, then dependabot will flag that all repos using it need to update it. So that was my plan.

PR #284, opened 2024-03-28
Create new :nearest strategy for Money::Allocator

raphpap:
see comment from @mkorostoff-shopify below

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ money_column expects a DECIMAL(21,3) database field.
 - Provides a `Money::Currency` class which encapsulates all information about a monetary unit.
 - Represents monetary values as decimals. No need to convert your amounts every time you use them. Easily understand the data in your DB.
 - Does NOT provide APIs for exchanging money from one currency to another.
-- Will not lose pennies during divisions
+- Will not lose pennies during divisions. For instance, given $1 / 3 the resulting chunks will be .34, .33, and .33. Notice that one chunk is larger than the others, so the result still adds to $1.
+- Allows callers to select a rounding strategy when dividing, to determine the order in which leftover pennies are given out.
 
 ## Installation
 
@@ -56,6 +57,30 @@ m.allocate([Rational(2, 3), Rational(1, 3)]).map(&:value) == [666.67, 333.33]
 ## Allocating up to a cutoff
 m.allocate_max_amounts([500, 300, 200]).map(&:value) == [500, 300, 200]
 m.allocate_max_amounts([500, 300, 300]).map(&:value) == [454.55, 272.73, 272.72]
+
+## Selectable rounding strategies during division
+
+# Assigns leftover subunits left to right
+m = Money::Allocator.new(Money.new(10.55, "USD"))
+monies = m.allocate([0.25, 0.5, 0.25], :roundrobin)
+#monies[0] == 2.64  <-- gets 1 penny
+#monies[1] == 5.28  <-- gets 1 penny
+#monies[2] == 2.63  <-- gets no penny
+
+# Assigns leftover subunits right to left
+m = Money::Allocator.new(Money.new(10.55, "USD"))
+monies = m.allocate([0.25, 0.5, 0.25], :roundrobin_reverse)
+#monies[0] == 2.63  <-- gets no penny
+#monies[1] == 5.28  <-- gets 1 penny
+#monies[2] == 2.64  <-- gets 1 penny
+
+# Assigns leftover subunits to the nearest whole subunit
+m = Money::Allocator.new(Money.new(10.55, "USD"))
+monies = m.allocate([0.25, 0.5, 0.25], :nearest)
+#monies[0] == 2.64  <-- gets 1 penny
+#monies[1] == 5.27  <-- gets no penny
+#monies[2] == 2.64  <-- gets 1 penny
+# $2.6375 is closer to the next whole penny than $5.275
 
 # Clamp
 Money.new(50, "USD").clamp(1, 100) == Money.new(50, "USD")
--- a/lib/money/allocator.rb
+++ b/lib/money/allocator.rb
@@ -9,18 +9,27 @@ class Money
 
     ONE = BigDecimal("1")
 
-    # Allocates money between different parties without losing pennies.
-    # After the mathematically split has been performed, left over pennies will
-    # be distributed round-robin amongst the parties. This means that parties
-    # listed first will likely receive more pennies than ones that are listed later
+    # Allocates money between different parties without losing subunits. A "subunit"
+    # in this context is the smallest unit of a currency that can be divided no
+    # further. In USD the unit is dollars and the subunit is cents. In JPY the unit
+    # is yen and the subunit is also yen. So given $1 divided by 3, the resulting subunits
+    # should be [34¢, 33¢, 33¢]. Notice that one of these chunks is larger than the other
+    # two, because we cannot transact in amounts less than 1 subunit.
+    #
+    # After the mathematically split has been performed, left over subunits will
+    # be distributed round-robin or nearest-subunit strategy amongst the parties.
+    # Round-robin strategy has the virtue of being easier to understand, while
+    # nearest-subunit is a more complex alogirthm that results in the most fair
+    # distribution.
     #
     # @param splits [Array<Numeric>]
     # @param strategy Symbol
     # @return [Array<Money>]
     #
     # Strategies:
-    # - `:roundrobin` (default): leftover pennies will be accumulated starting from the first allocation left to right
-    # - `:roundrobin_reverse`: leftover pennies will be accumulated starting from the last allocation right to left
+    # - `:roundrobin` (default): leftover subunits will be accumulated starting from the first allocation left to right
+    # - `:roundrobin_reverse`: leftover subunits will be accumulated starting from the last allocation right to left
+    # - `:nearest`: leftover subunits will by given first to the party closest to the next whole subunit
     #
     # @example
     #   Money.new(5, "USD").allocate([0.50, 0.25, 0.25])
@@ -38,29 +47,45 @@ class Money
     #   Money.new(30, "USD").allocate([Rational(2, 3), Rational(1, 3)])
     #     #=> [#<Money value:20.00 currency:USD>, #<Money value:10.00 currency:USD>]
 
-    # @example left over pennies distributed reverse order when using roundrobin_reverse strategy
+    # @example left over subunits distributed reverse order when using roundrobin_reverse strategy
     #   Money.new(10.01, "USD").allocate([0.5, 0.5], :roundrobin_reverse)
     #     #=> [#<Money value:5.00 currency:USD>, #<Money value:5.01 currency:USD>]
+
+    # @examples left over subunits distributed by nearest strategy
+    #   Money.new(10.55, "USD").allocate([0.25, 0.5, 0.25], :nearest)
+    #     #=> [#<Money value:2.64 currency:USD>, #<Money value:5.27 currency:USD>, #<Money value:2.64 currency:USD>]
+
     def allocate(splits, strategy = :roundrobin)
       splits.map!(&:to_r)
       allocations = splits.inject(0, :+)
 
       if (allocations - ONE) > Float::EPSILON
-        raise ArgumentError, "splits add to more than 100%"
+        raise ArgumentError, "allocations add to more than 100%"
       end
 
       amounts, left_over = amounts_from_splits(allocations, splits)
 
-      left_over.to_i.times do |i|
-        amounts[allocation_index_for(strategy, amounts.length, i)] += 1
+      order = case strategy
+      when :roundrobin
+        (0...left_over).to_a
+      when :roundrobin_reverse
+        (0...amounts.length).to_a.reverse
+      when :nearest
+        rank_by_nearest(amounts)
+      else
+        raise ArgumentError, "Invalid strategy. Valid options: :roundrobin, :roundrobin_reverse, :nearest"
       end
 
-      amounts.collect { |subunits| Money.from_subunits(subunits, currency) }
+      left_over.to_i.times do |i|
+        amounts[order[i]][:whole_subunits] += 1
+      end
+
+      amounts.map { |amount| Money.from_subunits(amount[:whole_subunits], currency) }
     end
 
     # Allocates money between different parties up to the maximum amounts specified.
-    # Left over pennies will be assigned round-robin up to the maximum specified.
-    # Pennies are dropped when the maximums are attained.
+    # Left over subunits will be assigned round-robin up to the maximum specified.
+    # Subunits are dropped when the maximums are attained.
     #
     # @example
     #   Money.new(30.75).allocate_max_amounts([Money.new(26), Money.new(4.75)])
@@ -90,6 +115,7 @@ class Money
       total_allocatable = [maximums_total.subunits, self.subunits].min
 
       subunits_amounts, left_over = amounts_from_splits(1, splits, total_allocatable)
+      subunits_amounts.map! { |amount| amount[:whole_subunits] }
 
       subunits_amounts.each_with_index do |amount, index|
         break unless left_over > 0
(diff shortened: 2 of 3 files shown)


lib/money/allocator.rb
```diff
@@ -119,10 +148,15 @@ def amounts_from_splits(allocations, splits, subunits_to_split = subunits)
...
+      amounts = []
+      splits.each_with_index do |ratio, i|
+        whole_pennies = (subunits_to_split * ratio / allocations.to_r).floor
+        fractional_pennies = (subunits_to_split * ratio / allocations.to_r).to_f - whole_pennies
+        left_over -= whole_pennies
+        amounts.push(
+          :whole_pennies => whole_pennies,
+          :fractional_pennies => fractional_pennies
+        )
       end
```
jenny-codes, line 160: nit: what's more idiomatic in Ruby is to use `.map`, so 

```ruby
splits.each_with_index.map do |ratio, i|
  ... (logic)
  {
     :whole_pennies => whole_pennies,
     :fractional_pennies => fractional_pennies
  }
end
```

lib/money/allocator.rb
```diff
@@ -50,12 +58,32 @@ def allocate(splits, strategy = :roundrobin)
...
 
       amounts, left_over = amounts_from_splits(allocations, splits)
+      result = []
 
-      left_over.to_i.times do |i|
-        amounts[allocation_index_for(strategy, amounts.length, i)] += 1
+      case strategy
+      when :roundrobin
```
jenny-codes, line 64: nit: more idiomatic in Ruby:

```ruby
result = case strategy
when ...
```

jenny-codes (approved):
Looks good! Left some non-blocking comments

lib/money/allocator.rb
```diff
@@ -51,11 +59,31 @@ def allocate(splits, strategy = :roundrobin)
...
+        end
+        right_to_left.reverse
+      when :nearest
+        nearest = amounts.map { |amount| amount[:whole_pennies] }
+        ranking = rank_by_nearest(amounts)
+        left_over.to_i.times do |i|
+          nearest[ranking[i]] += 1
+        end
+        nearest
+      else
```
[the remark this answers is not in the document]
mkorostoff-shopify: @raphpap I would be happy to implement that change, and I will do so today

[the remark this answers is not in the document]
mkorostoff-shopify: @raphpap thank you again for this feedback, I have implemented it as you described, and expanded test cases for all three allocation strategies to ensure that I have not inadvertently changed the output. I departed from your proposal only in one stylistic matter, introducing explicit ["first" and "last" variables](https://github.com/Shopify/money/blob/28de8b96c66c06e82217340d8e4c89e87ca3af16/lib/money/allocator.rb#L67-L68) to the roundrobin_reverse strategy, which I feel enhances readability, though I would be happy to adopt the single line style if you prefer.

lib/money/allocator.rb
```diff
@@ -51,11 +59,26 @@ def allocate(splits, strategy = :roundrobin)
...
 
+      order = case strategy
+      when :roundrobin
+        # returns [0,1,2 etc]
+        (0...left_over).to_a
+      when :roundrobin_reverse
+        first = amounts.length - left_over
+        last = amounts.length
+        # returns [5,4,3 etc]
+        (first...last).to_a.reverse
```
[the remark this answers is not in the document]
mkorostoff-shopify: Done. I had initially ruled out such a solution, because it results in the `order` variable being larger than the number of leftover pennies, e.g. you could have `order = [7, 6, 5, 4, 3, 2, 1]` and only 1 leftover penny. But thinking about it more now, I realize this does not actually matter because, the number of iterations is constrained on [line 73](https://github.com/Shopify/money/blob/469438a221118cef3be366c6845c2ac8df83bfcf/lib/money/allocator.rb#L73-L75).

spec/allocator_spec.rb
```diff
@@ -53,26 +53,99 @@
...
+    specify "#allocate raise ArgumentError when invalid strategy is provided" do
+      expect { new_allocator(0.03).allocate([0.5, 0.5], :bad_strategy_name) }.to raise_error(ArgumentError, "Invalid strategy. Valid options: :roundrobin, :roundrobin_reverse, :nearest")
     end
 
-    specify "#allocate fills pennies from end to beginning with roundrobin_reverse strategy" do
-      moneys = new_allocator(0.05).allocate([0.3,0.7], :roundrobin_reverse)
-      expect(moneys[0]).to eq(Money.new(0.01))
-      expect(moneys[1]).to eq(Money.new(0.04))
+    specify "#allocate raises an error when splits exceed 100%" do
+      expect { new_allocator(0.03).allocate([0.5, 0.6]) }.to raise_error(ArgumentError, "splits add to more than 100%")
```
raphpap, line 61: @mkorostoff-shopify are there tests in place where the split is under 100% ?

Does it work / how does it behave?

does [0.3,0.3] == [0.5,0.5]

I am currently wondering if _this_ could be when left_over.length might be longer than amounts.length 🤔

raphpap: Got it, thanks :+1:

Feel free to add a test if you think it's valuable.

I would also like to see some tests for non subunits=>decimal currencies (i.e. JPY which has no decimals), just in case 🤔

mkorostoff-shopify: @raphpap thank you for this feedback, I have implemented the tests you suggested for yen allocation as well as allocations adding to less than 100% in my latest commit, 803cb3f

lib/money/allocator.rb
```diff
@@ -132,15 +162,16 @@ def all_rational?(splits)
       splits.all? { |split| split.is_a?(Rational) }
     end
 
-    def allocation_index_for(strategy, length, idx)
-      case strategy
-      when :roundrobin
-        idx % length
-      when :roundrobin_reverse
-        length - (idx % length) - 1
-      else
```
raphpap, line 141: as a note - we didn't re-implement the `% length` in the `def allocate` above as it should be mathematically impossible to have more `left_over`s to re-distribute than we have `amounts.length`.

The `left_over` value should sum up to the amounts `fractionnal_subunits`

spec/allocator_spec.rb
```diff
@@ -53,27 +53,173 @@
       expect { new_allocator(1).allocate([rate, 1 - rate]) }.not_to raise_error
     end
 
+    specify "#allocate raise ArgumentError when invalid strategy is provided" do
+      expect { new_allocator(0.03).allocate([0.5, 0.5], :bad_strategy_name) }.to raise_error(ArgumentError, "Invalid strategy. Valid options: :roundrobin, :roundrobin_reverse, :nearest")
+    end
```
csalvato, line 58: Nice additional test coverage for existing behavior 🙏🏼

spec/allocator_spec.rb
```diff
@@ -53,27 +53,173 @@
...
+
+    specify "#allocate raises an error when splits exceed 100%" do
+      expect { new_allocator(0.03).allocate([0.5, 0.6]) }.to raise_error(ArgumentError, "splits add to more than 100%")
+    end
+
+    specify "#allocate scales up allocations less than 100%, preserving the relative magnitude of each chunk" do
+      # Allocations sum to 0.3
+      # This is analogous to new_allocator(12).allocate([1/3, 2/3])
+      moneys = new_allocator(12).allocate([0.1, 0.2])
```
csalvato, line 67: Nit: shouldn't this be `monies`?

csalvato: 🤷🏼 

https://www.perplexity.ai/search/Is-the-correct-xEGpmiD7SqWDCO086AOu5g#0

> ## Pluralization of "Money"
> 
> The correct plural form of the word "money" is **"monies"**.
> 
> Here are the key points:
> 
> ### "Monies" vs. "Moneys"
> 
> - **"Monies"** is the standard and most common plural form of "money". This is the preferred plural in most contexts.
> - **"Moneys"** is also an accepted plural form, but it is less common and generally considered more formal or technical.
> 
> ### Usage Guidelines
> 
> - Use "monies" when referring to sums or amounts of money, especially in a general or collective sense.
>   - Example: "The organization has received significant monies in donations this year."
> - Use "moneys" when referring to specific currencies or monetary systems.
>   - Example: "The different moneys used in the European Union include the euro, the British pound, and the Swiss franc."
> 
> So in summary, "monies" is the recommended and more widely used plural form of "money" in most everyday contexts. "Moneys" is a valid alternative, but it is less common and tends to be used in more formal or specialized situations.

spec/allocator_spec.rb
```diff
@@ -53,27 +53,173 @@
...
+      moneys = new_allocator(101).allocate([0.25, 0.25, 0.25, 0.25], :roundrobin)
+      expect(moneys[0]).to eq(Money.new(25.25)) # <-- gets no penny
+      expect(moneys[1]).to eq(Money.new(25.25)) # <-- gets no penny
+      expect(moneys[2]).to eq(Money.new(25.25)) # <-- gets no penny
+      expect(moneys[3]).to eq(Money.new(25.25)) # <-- gets no penny
     end
 
     specify "#allocate fills pennies from end to beginning with roundrobin_reverse strategy" do
+      #round robin reverse for 1 penny
       moneys = new_allocator(0.05).allocate([0.3,0.7], :roundrobin_reverse)
```
csalvato, line 119: Sidenote: I am kinda confused why the gem supports both. I probably just don't have context. But it'd be simpler for gem consumers to just reverse their array, wouldn't it?

```rb
      moneys = new_allocator(0.05).allocate([0.3,0.7].reverse)
```

Then the gem has one less thing to support.

spec/allocator_spec.rb
mkorostoff-shopify: thank you for the feedback, I will update to `monies`

spec/allocator_spec.rb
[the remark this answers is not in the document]
mkorostoff-shopify: thank you, that is very kind (FYI this is my first PR at shopify, so this feedback is especially welcomed)

spec/allocator_spec.rb
```diff
@@ -53,27 +53,173 @@
       expect { new_allocator(1).allocate([rate, 1 - rate]) }.not_to raise_error
     end
 
+    specify "#allocate raise ArgumentError when invalid strategy is provided" do
+      expect { new_allocator(0.03).allocate([0.5, 0.5], :bad_strategy_name) }.to raise_error(ArgumentError, "Invalid strategy. Valid options: :roundrobin, :roundrobin_reverse, :nearest")
+    end
+
+    specify "#allocate raises an error when splits exceed 100%" do
+      expect { new_allocator(0.03).allocate([0.5, 0.6]) }.to raise_error(ArgumentError, "splits add to more than 100%")
```
elfassy, line 61: Money has another method called `split`, I think we should avoid that terminology here
```suggestion
      expect { new_allocator(0.03).allocate([0.5, 0.6]) }.to raise_error(ArgumentError, "allocations add to more than 100%")
```

spec/allocator_spec.rb
mkorostoff-shopify: fixed

spec/allocator_spec.rb
```diff
@@ -53,27 +53,173 @@
...
+      expect(moneys[1]).to eq(Money.new(25.25)) # <-- gets no penny
+      expect(moneys[2]).to eq(Money.new(25.25)) # <-- gets no penny
+      expect(moneys[3]).to eq(Money.new(25.25)) # <-- gets no penny
     end
 
-    specify "#allocate does not raise ArgumentError when invalid splits types are provided" do
-      moneys = new_allocator(0.03).allocate([0.5, 0.5], :roundrobin)
-      expect(moneys[0]).to eq(Money.new(0.02))
-      expect(moneys[1]).to eq(Money.new(0.01))
+    specify "#allocate :roundrobin strategy distributes leftover Yen from left to right" do
```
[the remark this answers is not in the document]
mkorostoff-shopify: updated, thanks for the feedback

spec/allocator_spec.rb
csalvato: What do you think @elfassy ?

elfassy: my memory is a bit blurry, but I think there was a valid reason. Would need to have a closer look (on monday). Here's the original PR https://github.com/Shopify/money/pull/186

lib/money/allocator.rb
```diff
@@ -132,15 +162,16 @@ def all_rational?(splits)
...
+    # For instance, given inputs [1.1, 1.5, 1.9] the correct ranking is 2, 1, 0. This is because 1.9 is nearly 2.
+    # Note that we are not ranking by absolute size, we only care about the distance between our input number and
+    # the next whole number. Similarly, given the input [9.1, 5.5, 3.9] the correct ranking is *still* 2, 1, 0. This
+    # is because 3.9 is nearer to 4 than 9.1 is to 10.
+    def rank_by_nearest(amounts)
+      nearest = amounts.map.with_index  do |amount, i|
+        { :index => i, :distance => 1 - amount[:fractional_subunits] }
       end
+      nearest.sort_by { |n| n[:distance] }.map { |n| n[:index] }
     end
```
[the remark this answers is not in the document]
csalvato: Actually, according to Phind, it will only iterate over the array once :scream:

https://www.phind.com/search?cache=k6rzfykizt2xp1ohjz6x374u

README.md
```diff
@@ -11,7 +11,8 @@ money_column expects a DECIMAL(21,3) database field.
 - Provides a `Money::Currency` class which encapsulates all information about a monetary unit.
 - Represents monetary values as decimals. No need to convert your amounts every time you use them. Easily understand the data in your DB.
 - Does NOT provide APIs for exchanging money from one currency to another.
-- Will not lose pennies during divisions
+- Will not lose pennies during divisions. For instance, given $1 / 3 the resulting chunks will be .34, .33, and .33. Notice that one chunk is larger than the others, so the result still adds to $1.
```
csalvato, line 14: Useful callout 👍🏼

README.md
```diff
@@ -11,7 +11,8 @@ money_column expects a DECIMAL(21,3) database field.
...
+- Allows callers to select a rounding strategy when dividing, to determine the order in which leftover pennies are given out.
```
csalvato, line 15: ❤️

csalvato (approved):

lib/money/allocator.rb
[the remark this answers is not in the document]
csalvato: Benchmarked with this code:

```rb
require 'benchmark/ips'

Benchmark.ips do |x|
  # Configure the number of seconds used during
  # the warmup phase (default 2) and calculation phase (default 5)
  x.config(:time => 5, :warmup => 2)

  amounts = Array.new(4) { { whole_subunits: rand(100), fractional_subunits: rand } }


  x.report("one-liner") do
    amounts.each_with_index.sort_by{ |amount, _| 1 - amount[:fractional_subunits] }.map(&:last)
  end

  # Typical mode, runs the block as many times as it can
  x.report("original") do
    nearest = amounts.map.with_index  do |amount, i|
      { :index => i, :distance => 1 - amount[:fractional_subunits] }
    end
    nearest.sort_by { |n| n[:distance] }.map { |n| n[:index] }
  end

  # Compare the iterations per second of the various reports!
  x.compare!
end
```

When there are 4 elements in the array, the original is 1.04x slower:

```
           one-liner    43.185k i/100ms
            original    42.775k i/100ms
           one-liner    440.940k (± 1.5%) i/s -      2.246M in   5.094009s
            original    422.052k (± 2.6%) i/s -      2.139M in   5.071319s

Comparison:
           one-liner:   440940.3 i/s
            original:   422052.3 i/s - 1.04x  slower
```

When there are 1000 elements in the array, the original is 1.28x slower:

```
           one-liner   275.000 i/100ms
            original   217.000 i/100ms
           one-liner      2.802k (± 0.8%) i/s -     14.025k in   5.004871s
            original      2.189k (± 0.9%) i/s -     11.067k in   5.056980s

Comparison:
           one-liner:     2802.4 i/s
            original:     2188.7 i/s - 1.28x  slower
```

So the one-liner that uses the enumerators is fast enough to use the implementation that avoids 3 iterations 🎉

PR #287, opened 2024-04-05
Optimize `Money#split(n)` for large `n`

casperisfine:
When passing a very large number to split, it result in allocating between 2 and 3 arrays of such size. e..g `Money.new(1_000_000).split(1_000_000)` allocates an array of size 1M, then concat it with an empty array, which cause Ruby to allocate another array of 1M entries.

All this cause two `8MB` mallocs, which doesn't sound like that much but if done a few times is likely to trigger Ruby GC.

Instead we can return a small enumerable that will simply act like an array, but won't actually materialize it.


lib/money/money.rb
```diff
@@ -34,6 +34,96 @@ def *(other)
     end
   end
 
+  class SplitList
```
elfassy, line 37: I'd suggest moving this concern to a new `money/splitter.rb` file, similar to https://github.com/Shopify/money/blob/main/lib/money/allocator.rb. 
This class could have two public methods `split` and `calculate_split`.

casperisfine: Done.

lib/money/splitter.rb
```diff
@@ -0,0 +1,109 @@
...
+      raise ArgumentError, "need at least one party" if num < 1
+      subunits = money.subunits
+      low = Money.from_subunits(subunits / num, money.currency)
+      high = Money.from_subunits(low.subunits + 1, money.currency)
+
+      num_high = subunits % num
+
+      @split = {}
+      @split[high] = num_high if num_high > 0
+      @split[low] = num - num_high
```
elfassy, line 21: I'd recommend moving this out of the initializer. We only need to compute this once it's used. It could also probably prevent us from having to `dup` above, by creating a new hash (although with the freeze, I'm not sure if dup is needed at all)

lib/money/money.rb
```diff
@@ -316,17 +316,7 @@ def split(num)
...
-    low = Money.from_subunits(subunits / num, currency)
-    high = Money.from_subunits(low.subunits + 1, currency)
-
-    num_high = subunits % num
-
-    {}.tap do |result|
-      result[high] = num_high if num_high > 0
-      result[low] = num - num_high
-    end
+    Splitter.new(self, num).split.dup
```
elfassy, line 319: What do you think of the idea of being able to call either `to_h` or `to_a` on this Splitter object depending on the format you want? 
```suggestion
    Splitter.new(self, num).to_h
```
to me it's easier to reason about than `split`, which is unclear if it's an array. Long term we could even deprecate this `calculate_splits` method all together in favor of `money.split(3).to_h`

casperisfine: This isn't a proper implementation for `to_h`. This object is an `Enumerable` that replaces an Array and tries to act as much like an Array as possible. 

If you define `to_h` like this you'll break lots of code and assumptions. `Splitter#to_h` should behave like `Array#to_h`.

lib/money/splitter.rb
casperisfine: > There will likely be cases where this object is created but never used

I doubt so, but ok.

elfassy (approved):

PR #289, opened 2024-04-11
Revert "Drop support for initialization with another money object."

elfassy:
Reverts Shopify/money#279

# Why

Removing the ability from Money.new to accept a money object is creating toil on multiple repos, including core, preventing the gem to be updated
ex: https://github.com/Shopify/shipify/pull/27262

It will also allow us to stay consistent with expectations shared with the RubyMoney gem https://github.com/RubyMoney/money/blob/main/spec/money_spec.rb#L57-L60

We still want to prevent changing the currency as mentioned in the original PR description.  

# What

- allow Money.new to accept a money object (revert)
- raise if Money.new is passed a money object with a different currency, that would cause the currency to change


mkorostoff-shopify:
Looks good to me! The one thing I'm uncertain about is how/where the version numbering should be configured. This would would go in 2.2.0 (not 2.1.0 as shown in the lockfile) but we could make a separate PR for that.

mkorostoff-shopify (approved):

PR #283, opened 2024-03-22
improve the wording of the to_money deprecations warning

elfassy:
## Why

We want to track deprecations happening with the `.to_money` method seperatly from those from arithmetic computations. This will allow us to safely raise knowing these issues no longer happen in production

## What

Introduce a new deprecation warning message for `to_money` with inconsistent currencies

Before:
```
Money.new(1, "USD").to_money("CAD") 
# "mathematical operation not permitted for Money objects with different currencies CAD and USD. A Money::IncompatibleCurrencyError will raise in the next major release"
```

After:
```
Money.new(1, "USD").to_money("CAD") 
# "to_money is attempting to change currency of an existing money object from USD to CAD. A Money::IncompatibleCurrencyError will raise in the next major release"
```

## Note

This only affects those using the `legacy_deprecations` option. Others will continue to get an `Money::IncompatibleCurrencyError` exception

@Shopify/shopify-money

PR #294, opened 2024-04-24
support coercible objects that respond to `to_money`

elfassy:
# Why

Some apps have their own class of objects that used to be able to be coerced into money objects (ex: MoneyBag). This brings back the legacy behaviour that had been modified in https://github.com/Shopify/money/pull/283

```ruby
class Foo
  attr_accessor :amount
  def to_money(*)
      Money.new(amount)
   end
end

Money.new(1) > Foo.new(amount: 1) # false
```

# What

For code bases using the `legacy_deprecation` option only:
- call `to_money` on the object being coerced into a money
- show a deprecation warning

For everyone else:
- we continue raising an exception. Apps should convert to money objects on their own, ex:
```ruby
Money.new(1) > Foo.new(amount: 1).to_money # false
```

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -224,14 +224,8 @@ class Money
       return Money.new(value, new_currency)
     end
 
-    unless currency.compatible?(Helpers.value_to_currency(new_currency))
-      msg = "to_money is attempting to change currency of an existing money object from #{currency} to #{new_currency}"
-      if Money.config.legacy_deprecations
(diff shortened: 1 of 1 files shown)


mkorostoff-shopify (approved):
LGTM!

elfassy:
I'm on the fence about shipping this... What do we expect `Foo.new(1) + Money.new(1)` class to be? A Money or of Foo (right now it'll be a Money)? This means that if someone does something like `foos.sum(Money.new(0))` they could unknowingly convert the answer to a Money object. If we ship this, we should deprecate this behaviour

PR #301, opened 2024-05-06
remove legacy thousand fuzzy parsing

elfassy:
# Why

For legacy reasons the FuzzyParser would convert `1.000` to `1_000` irregardless of currency. This is no longer required

# What

Apps using this parser will convert `1.000` using the delimiter of the currency

```ruby
Money::Parser::Fuzzy.parse("1.000", "USD") == Money.new(1, "USD")
Money::Parser::Fuzzy.parse("1.000", "EUR") == Money.new(1000, "EUR")
```

### DIFF
--- a/lib/money/parser/fuzzy.rb
+++ b/lib/money/parser/fuzzy.rb
@@ -161,11 +161,6 @@ class Money
           return true
         end
 
-        # legacy support for 1.000 USD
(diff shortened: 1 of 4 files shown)


elfassy:
closing in favor of a deprecation warning. It's unclear if this is a feature or a bug yet... we need to collect more info

PR #305, opened 2024-06-07
[NullCurrency] Set minor units to 4

kellydanma:
Part of https://github.com/Shopify/shopify/issues/511351

## ❓ Why this fix

Our GraphQL scalar `Money` currently rounds all amounts with > 2 decimal places. This causes incorrect refund amount, as described in this comment: https://github.com/Shopify/shopify/issues/511351#issuecomment-2159230147

When money values are rounded incorrectly, our merchants suffer. 

## 💥 Impact of this PR

This will set the default currency to 4 minor units, which means the subunit to unit will be 10000 instead of 100. Are there any breaking features I should be aware of? I'm fairly certain this is considered **breaking** according to [semantic versioning](https://semver.org/).

### DIFF
--- a/lib/money/null_currency.rb
+++ b/lib/money/null_currency.rb
@@ -44,8 +44,8 @@ class Money
       @iso_numeric           = '999'
       @name                  = 'No Currency'
       @smallest_denomination = 1
-      @subunit_to_unit       = 100
-      @minor_units           = 2
+      @subunit_to_unit       = 10000
+      @minor_units           = 4
       @decimal_mark          = '.'
       freeze
     end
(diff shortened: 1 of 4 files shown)


kellydanma:
👋 @elfassy Hi! Before I put more work into this PR, can you give it some 👀 and let me know if this fix is okay? Are there any historic reasons why `NULL_CURRENCY` is 2 decimals places only?

elfassy:
yes, we want NullCurrency to behave like USD for legacy apps who migrated from pre-v1 version of this gem (before money objects accepted a currency). If you wish to store more than two decimal accuracy, it's likely you want to keep values and currencies separate

kellydanma:
>

kellydanma:
@elfassy Just one more question before we close this. I responded in https://github.com/Shopify/shopify/issues/511351#issuecomment-2159588854 🙏

PR #307, opened 2024-06-27
Remove deprecation notice and raise error when strings cannot be parsed by bigdecimal

TayKangSheng:
issue: https://github.com/Shopify/money/pull/307

# Why

Currently we attempt to parse strings with BigDecimal, and we default to `BigDecimal(0)` for cases where the string cannot be parsed. In v3, we want to remove this fallback and we recommend users to parse the money first before using initializing a new money object. 

# What

**This is a breaking change**

We now raise ArgumentError if the provided value is a string and cannot be parsed by BigDecimal. 

```ruby
Money.new("$1.00", "USD") # -> raises invalid value for BigDecimal(): "$1.00" (ArgumentError)
```

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -29,11 +29,7 @@ class Money
         when Rational
           BigDecimal(num, MAX_DECIMAL)
         when String
-          decimal = BigDecimal(num, exception: !Money.config.legacy_deprecations)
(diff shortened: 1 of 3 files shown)


elfassy (approved):
last event in the logs was 4 weeks ago, we're good to merge this in v3!

PR #309, opened 2024-06-27
to_money parse with Money.new when there is no thousand delimiter

elfassy:
# Why

Make `"1.123".to_money` behave consistently like `Money.new(1.123)` and have a value of `1.12`

# What

Remove deprecation warning for apps using `legacy_deprecation` option, and default to the correct behaviour

Note: This is a BREAKING CHANGE for apps using this option!

### DIFF
--- a/lib/money/core_extensions.rb
+++ b/lib/money/core_extensions.rb
@@ -20,18 +20,18 @@ class String
       return Money.new(self, currency)
     end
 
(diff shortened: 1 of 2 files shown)


elfassy:
core is ok with this change https://github.com/Shopify/shopify/issues/517062#issuecomment-2195231200

TayKangSheng (approved):
LGTM

PR #310, opened 2024-06-27
raise UnknownCurrency error when currency is invalid

TayKangSheng:
Issue: https://github.com/Shopify/shopify/issues/517112

# Why 

Support for invalid currency code has been deprecated and we want to remove support in V3. 

# What

Raise error when currency code is invalid when initializing a money object including reading from database.


elfassy (changes requested):
not ready to ship this yet

PR #311, opened 2024-06-27
convert currency

elfassy:
# Why

`Money.new(money * exchange_rate, "JPY")` will raise an exception since `money * exchange_rate` returns a money object. A valid alternative is `Money.new(money.value * exchange_rate, "JPY")`

to avoid confusion, this PR provides a nice helper method that should be the prefered "explicit" way of doing currency conversions

# What

```ruby
money.convert_currency(exchange_rate, "JPY")
```


TayKangSheng (approved):
👍

PR #312, opened 2024-06-27
raise error when comparing or casting to a different currency

TayKangSheng:
Issue: https://github.com/Shopify/shopify/issues/517120, https://github.com/Shopify/core-issues/issues/87436

# Why 

Removing support for
1. mathematical operation with different currencies
2. Casting the currency of an existing money object with currency to another currency

# What

Raise error if any of the above happens


elfassy (changes requested):
not ready to ship this yet

PR #313, opened 2024-06-27
arithmetic raises if unable to coerce to money

elfassy:
# Why

Removing deprecations no longer needed for v3

# What

BREAKING CHANGE for people with the `legacy_deprecation` option enabled

We no longer try to coerce inside the gem by calling `to_money`. You need to do it on your own

```ruby
# before
your_own_object + Money.new(1)

# after
your_own_object.to_money + Money.new(1)
```

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -371,12 +371,7 @@ class Money
       yield(Money.new(other, currency))
 
     else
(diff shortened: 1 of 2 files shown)


elfassy:
merged here https://github.com/Shopify/money/pull/314

PR #316, opened 2024-06-28
Improve Fuzzy parsing with invalid strings

elfassy:
# Why

The fuzzy parsing tries to guess the amount from a string. We can get a little more strict about the guesses we're making

# What

- BREAKING CHANGE: In cases where it's not able to make a guess at all, we should return `nil` instead of `0`


TayKangSheng (approved):
Okay. With this change, we will need to check all the places that calls Fuzzy.parse to check for `nil` value and handle with a fallback when necessary.

elfassy:
I'm going to split the `nil` behaviour in it's own PR. I'd still like to get the other changes in v3

PR #317, opened 2024-07-09
Support a "money string" format with `.to_money_string` and `Money.from_money_string`

shawnfrench:
Looking for comments (and/or rude gestures) on adding a `money string` concept to promote simple/standard string representation of money with both amount and currency. 

Why? JSON and YAML (the currently supported formats) are great for some things, but for others (simple and efficient transport of money details) not so much.

I propose a string in the format of `<currency> <amount>` as the simplest most universal way to combine the two (with a space in-between) and have added 2 methods to do so along with entries to the readme for example usage.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,12 @@ Money.new(50, "USD").clamp(1, 100) == Money.new(50, "USD")
 Money.from_subunits(500, "USD")  == Money.new(5, "USD")   # 5 USD
 Money.from_subunits(5, "JPY")    == Money.new(5, "JPY")   # 5 JPY
 Money.from_subunits(5000, "TND") == Money.new(5, "TND")   # 5 TND
+
+# String formatting
+m = Money.new(12.34, "CAD")
+m.to_s                                #=> "12.34"
+m.to_money_string                     #=> "CAD 12.34"
+Money.from_money_string("CAD 12.34")  #=> #<Money value:12.34 currency:CAD>
 ```
 
 ## Currency
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -74,6 +74,11 @@ class Money
       new(value, currency)
     end
 
+    def from_money_string(ms)
+      currency, value = ms.split(' ', 2)
+      new(value, currency)
+    end
+
     def rational(money1, money2)
       money1.send(:arithmetic, money2) do
         factor = money1.currency.subunit_to_unit * money2.currency.subunit_to_unit
@@ -258,6 +263,10 @@ class Money
   end
   alias_method :to_s, :to_fs
   alias_method :to_formatted_s, :to_fs
+
+  def to_money_string
+    "#{currency.to_s} #{to_s(:amount)}"
+  end
 
   def to_json(options = nil)
     if (options.is_a?(Hash) && options.delete(:legacy_format)) || Money.config.legacy_json_format
--- a/spec/money_spec.rb
+++ b/spec/money_spec.rb
@@ -62,6 +62,14 @@ RSpec.describe "Money" do
     expect{ Money.new(1, 'USD').to_money('CAD') }.to raise_error(Money::IncompatibleCurrencyError)
   end
 
+  it ".from_money_string" do
+    expect(Money.from_money_string("CAD 12.34")).to eq(Money.new(12.34, 'CAD'))
+  end
+
+  it "#to_money_string" do
+    expect(Money.new(12.34, 'CAD').to_money_string).to eq("CAD 12.34")
+  end
+
   it "defaults to 0 when constructed with no arguments" do
     expect(Money.new).to eq(Money.new(0))
   end


AnotherJoSmith:
I checked if the other Money library (https://github.com/RubyMoney/money) solved this in any way, and the only thing I saw that was somewhat close to this is [`.hash`](https://www.rubydoc.info/gems/money/Money#hash-instance_method) which converts the currency code to an integer, which has the disadvantage of not being human readable. So I think this looks fine to me.

In terms of API, I wonder if we should just call this `.serialize` and `.deserialize`, of if that's too generic? Don't have strong opinions on it.

lib/money/money.rb
```diff
@@ -259,6 +264,10 @@ def to_fs(style = nil)
   alias_method :to_s, :to_fs
   alias_method :to_formatted_s, :to_fs
 
+  def to_money_string
+    "#{currency.to_s} #{to_s(:amount)}"
```
bradly, line 268: I believe the `to_s` on `currency` is redundant here.
```suggestion
    "#{currency} #{to_s(:amount)}"
```

shawnfrench: I think it's required:
```
irb(main):001> m = Money.new(56.78, 'SEK')
=> #<Money value:56.78 currency:SEK>
irb(main):002> m.currency
=>
#<Money::Currency:0x000000011ff32800
 @decimal_mark=",",
 @disambiguate_symbol="SEK",
 @iso_code="SEK",
 @iso_numeric="752",
 @minor_units=2,
 @name="Swedish Krona",
 @smallest_denomination=100,
 @subunit_symbol=nil,
 @subunit_to_unit=100,
 @symbol="kr">
irb(main):003> m.currency.to_s
=> "SEK"
```

bradly: Since you are already inside a string you get it for free :smile:. I'm using a version this method until this is merged and Rubocop complained about the additional .to_s inside the interpolated string.

```
irb(main):002> m = Money.new(56.78, 'SEK')
=> #<Money value:56.78 currency:SEK>
irb(main):003> m.currency
=> #<Money::Currency:0x0000000100239118 @decimal_mark=",", @disambiguate_symbol="SEK", @iso_code="SEK", @iso_numeric="752", @minor_units=2, @name="Swedish Krona", @smallest_denomination=100, @subunit_symbol=nil, @subunit_to_unit=100, @symbol="kr">
irb(main):004> "#{m.currency}"
=> "SEK"
irb(main):005> "#{m.currency.to_s}"
=> "SEK"
```

csalvato:
## Strong opinion, loosely held

I don't think this sort of transform is a responsibility of this gem. 

It adds an unnecessary method to the interface that only applies for a narrow set of use cases based only on Shopify's specific needs and not something necessary in the OSS ecosystem.

## Rationale

If you want some service to receive strings that represent some money value, then the docs for that service can simply say the format that is expected (e.g. `<currency> <value>`).

Such a service should validate the inputs to ensure proper formatting anyway, so this would have to be documented.

Thus, in those docs, just include an example of how to convert `Money` into the string with the right format using vanilla rails:

```rb
ten_bucks = Money.new(10,"USD") => #<Money value:10.00 currency:USD>

# in Rails
number_to_currency(ten_bucks.value, unit: ten_bucks.currency.iso_code, format: "%u %n") => "USD 10.00"

# vanilla ruby
" #{ten_bucks.currency.iso_code} #{ten_bucks.value}"
```

IMO we don't need to make any changes to the gem to support that translation.

fredericboivin:
There's no such thing as "universal" when it comes to presenting money. Currency can be on the right or the left, significant digits varies, commas vs dots, etc. This isn't accurate for presenting amounts, it's less precise than JSON objects for transport and it shouldn't be used for any kind of calculations 😅

Considering that APIs communicate through JSON either way, that the payloads will be parsed into Money object before we do anything with them, I don't really see the use case. It's not simpler than JSON, it's just different. 

Arguably it _adds_ complexity because systems need to be aware of this "format" and identify it correctly in order to parse it into a Money object: if you see "12,99 CAD" or "$12.99 CAD" it's not a Money object, "12.99 CAD" could be but the currency is misplaced, is "1299 CAD" referring to 1299 cents, or 1299 dollars? There's endless edge cases to account for if we were to use this as transport format, which makes it strictly worse than JSON

tartakynov (commented):
From my experience, in Billing we found it's best not to persist amount and currency in the same field. 

Storing them separately makes stuff like summing, grouping, and filtering easier.

That's why billing keeps the amount and currency separate when serialized. Mixing them only makes sense when presenting to the customer on the front-end.

I think this PR makes it just too easy to store everything serialized in the same field and it will bite you later.

README.md
```diff
@@ -89,6 +89,12 @@ Money.new(50, "USD").clamp(1, 100) == Money.new(50, "USD")
 Money.from_subunits(500, "USD")  == Money.new(5, "USD")   # 5 USD
 Money.from_subunits(5, "JPY")    == Money.new(5, "JPY")   # 5 JPY
 Money.from_subunits(5000, "TND") == Money.new(5, "TND")   # 5 TND
+
+# String formatting
+m = Money.new(12.34, "CAD")
+m.to_s                                #=> "12.34"
+m.to_money_string                     #=> "CAD 12.34"
+Money.from_money_string("CAD 12.34")  #=> #<Money value:12.34 currency:CAD>
```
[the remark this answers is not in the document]
elfassy: I agree with @csalvato, it's not really the responsibility of this gem. RubyMoney moved their parsing logic to another gem (https://github.com/RubyMoney/monetize). We likely also want to remove parsers from this gem and move them to a new gem in the future

shawnfrench:
Apologies forgot about this - and thanks for all the insight into this codebase and our approach to serializing money. We'll just do this outside the gem cheers!

PR #326, opened 2024-08-13
Introduce from_json as the inverse of to_json

pedropb:
The Money gem provides two JSON generating methods:
- `to_json` generates a JSON-encoded String representation of the Money object.
- `as_json` generates a JSON-encoded Hash representation of the Money object.

This PR introduces a class method that accepts any of these 2 JSON representations and generates a Money object from it.

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -280,6 +280,7 @@ class Money
       { value: to_s(:amount), currency: currency.to_s }
     end
   end
+  alias_method :to_h, :as_json
 
   def abs
     abs = value.abs
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'forwardable'
+require 'json'
 
 class Money
   include Comparable
@@ -72,6 +73,11 @@ class Money
 
       value = Helpers.value_to_decimal(subunits) / subunit_to_unit_value
       new(value, currency)
+    end
+
+    def from_json(string)
+      hash = JSON.parse(string, symbolize_names: true)
+      Money.new(hash.fetch(:value), hash.fetch(:currency))
     end
 
     def rational(money1, money2)
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -80,6 +80,11 @@ class Money
       Money.new(hash.fetch(:value), hash.fetch(:currency))
     end
 
+    def from_hash(hash)
+      hash = hash.transform_keys(&:to_sym)
+      Money.new(hash.fetch(:value), hash.fetch(:currency))
+    end
+
     def rational(money1, money2)
       money1.send(:arithmetic, money2) do
         factor = money1.currency.subunit_to_unit * money2.currency.subunit_to_unit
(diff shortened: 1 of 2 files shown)


lib/money/money.rb
```diff
@@ -74,6 +75,16 @@ def from_subunits(subunits, currency_iso, format: :iso4217)
       new(value, currency)
     end
 
+  def from_json(hash_or_string)
```
mkorostoff-shopify, line 78: I wonder if this might be a bit easier to follow with explicit, separate methods `from_hash` and `from_json` that each handle a single data type, rather than attempting to juggle types within a single method. It seems a slightly surprising result to have a method from_json that accepts an input which is not JSON.

spec/money_spec.rb
```diff
@@ -388,6 +388,20 @@
     expect(JSON.dump(Money.new(1.00, "CAD"))).to eq('{"value":"1.00","currency":"CAD"}')
   end
 
+  it "creates Money object from JSON Hash in from_json" do
+    one_cad = Money.new(1, "CAD")
+    expect(Money.from_json(one_cad.as_json)).to eq(one_cad)
```
mkorostoff-shopify, line 393: It might be worthwhile to have an additional test case where the input is spelled out more explicitly, more as documentation than to guard against breaking changes, e.g.

```ruby
one_cad = Money.new(1, "CAD")
input = '{"amount": 1, "currency": "CAD"}'
expect(Money.from_json(input)).to eq(one_cad)
```

mkorostoff-shopify (commented):
overall I have no problem with these changes, just a couple of optional suggestions

lib/money/money.rb
elfassy: Do we need  `from_hash` at all? I propose only starting with `from_json`. Going from a hash to a Money object is trivial `Money.new(hash[:amount], hash[:currency])`

elfassy:
@pedropb if you want to add just `from_json` I can include it in the v3 release

pedropb:
Hey folks,

pushed some rebased commits. Here are the changes:
- as requested: separate methods to handle distinct types `from_json` and `from_hash`.
- new: aliased `to_h` as `as_json`
- Even though it is trivial to create a Money object from a Hash, I kept the `from_hash` for symmetry (inverse of `as_json`). This is the use case that we currently have in our application with direct access to `hash["value"], hash["currency"]` and would like to get rid of.

lib/money/money.rb
```diff
@@ -74,6 +75,16 @@ def from_subunits(subunits, currency_iso, format: :iso4217)
...
 
+    def from_hash(hash)
+      hash = hash.transform_keys(&:to_sym)
+      Money.new(hash.fetch(:value), hash.fetch(:currency))
+    end
+
+  def from_json(string)
+    hash = JSON.parse(string, symbolize_names: true)
+    Money.new(hash.fetch(:value), hash.fetch(:currency))
+  end
```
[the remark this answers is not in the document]
pedropb: Good 👀, thanks !
Pushed a fixed commit.

rosew (approved):
LGTM

PR #327, opened 2024-08-15
default to the currency for thousand delimiter when unknown

elfassy:
# Why

The fuzzy parsing tries to guess the amount from a string. We can get a little more strict about the guesses we're making

# What

When we're not able to guess if the delimiter is for thousands or decimals, defer to the currency decimal delimiter

### DIFF
--- a/lib/money/parser/fuzzy.rb
+++ b/lib/money/parser/fuzzy.rb
@@ -158,11 +158,6 @@ class Money
           return true
         end
 
-        # legacy support for 1.000 USD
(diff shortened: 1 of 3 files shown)


TayKangSheng (approved):
Theres a whole lot of commits in this PR. If i'm understanding correctly, the fix is in lib/money/core_extensions.rb where we call `Money.new` instead of running it through `Parser::Fuzzy`? 👍 Good on that change.

elfassy:
sorry the rebase on v3 caused a bit of noise on this PR, please have another look 🙏

PR #350, opened 2024-11-28
Add USDC stablecoin to `currency_non_iso.yml`

davidpatters0n:
### What is being done?
We (Post purchase order management) recently encountered the following exception: https://github.com/Shopify/shopify/issues/557605 TL;DR - We've imported a balance transaction into Shopify Payments from Stripe that has a currency of USDC. It seems when we call the Shopify money gem it throws `[Shopify/Money] Invalid iso4217 currency 'USDC' `. 

### Why is it being done?
We need to add this currency to support surfacing USDC transactions in Shopify properly. Also there is a initiative in the future for Shopify Payments to support USDC stablecoin.

### DIFF
--- a/config/currency_non_iso.yml
+++ b/config/currency_non_iso.yml
@@ -102,7 +102,7 @@ usdc:
   disambiguate_symbol: USDC
   alternate_symbols: []
   subunit: Cent
-  subunit_to_unit: 1000000,
+  subunit_to_unit: 100,
   symbol_first: false
   html_entity: "$"
   decimal_mark: "."


config/currency_non_iso.yml
```diff
@@ -94,3 +94,18 @@ gbx:
   thousands_separator: ","
   iso_numeric: ''
   smallest_denomination: 1
+usdc:
```
davidpatters0n, line 97: Contents of this file was taken from: https://github.com/RubyMoney/money/blob/main/config/currency_non_iso.json#L130

config/currency_non_iso.yml
```diff
@@ -94,3 +94,18 @@ gbx:
...
+  priority: 100
+  so_code: USDC
+  name: USD Coin
+  symbol: USDC
+  disambiguate_symbol: USDC
+  alternate_symbols: []
+  subunit: Cent
+  subunit_to_unit: 1000000,
```
robinbrandt, line 105: I think this brings up the discussion of the guidance around storing monetary values in the database. At the moment we are using `DECIMAL(21,3)`.

This will not be enough to represent this without a loss of precision.

cc @csalvato - I think you had brought up the topic before.

elfassy: @robinbrandt a missing currency has `subunit_to_unit: 100` so this would be consistent with the existing behaviour

davidpatters0n: yup that's right the above PR is where I got this information from. 👍🏽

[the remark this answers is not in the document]
davidpatters0n: do we want this to be different to the Ruby money gem ?

PR #360, opened 2025-02-05
no subunits for HUF

elfassy:
https://github.com/RubyMoney/money/issues/677#issue-202864984

PayPal agrees with this and does not support decimals (will return an error if provided) https://developer.paypal.com/reference/currency-codes/

[Stripe](https://docs.stripe.com/currencies) also says the same
Screenshot 2025-02-25 at 2 57 20 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ doc
 # jeweler generated
 pkg
 
-# Have editor/IDE/OS specific files you need to ignore? Consider using a global gitignore: 
+# Have editor/IDE/OS specific files you need to ignore? Consider using a global gitignore:
 #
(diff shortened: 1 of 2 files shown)


matid:
@elfassy HUF continues to be subdivided into 100 units in ISO 4217. Since the docs say `iso4217 (default) - Uses the standard ISO 4217 subunit definitions` I believe it should remain aligned to the actual published standard, not what Stripe/PayPal says.

Merging this has potentially serious consequences for anyone relying on this gem aligning to the standard when communicating with external banking providers using subdivision units.

Sources:
- https://en.wikipedia.org/wiki/ISO_4217
- https://datahub.io/core/currency-codes#codes-all

PR #361, opened 2025-02-05
no subunits for TWD 

elfassy:
# Why

While the ISO-4217 spec says that the currency still has decimals, it does look like this is no longer used outside of very few cases https://www.reddit.com/r/taiwan/comments/qfuoh8/question_about_twd_taiwan_currency_are_sub_dollar/

PayPal will give an error https://developer.paypal.com/reference/currency-codes/ if a subunit is passed when dealing with TWD

[Stripe](https://docs.stripe.com/currencies) also treats TWD the same (although still expects 00 to be passed)

Screenshot 2025-02-25 at 2 56 00 PM

# What

`subunit_to_unit: 1` which means going forward, any amount with decimals like `Money.new(1.23, "TWD")` will be rounded to equal `Money.new(1, "TWD")`

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -12,6 +12,7 @@ class Money
     STRIPE_SUBUNIT_OVERRIDE = {
       'ISK' => 100,
       'UGX' => 100,
+      'TWD' => 100,
     }.freeze
 
     def value_to_decimal(num)


elfassy:
needs a rebase now that https://github.com/Shopify/money/pull/404 shipped

matid:
> While the ISO-4217 spec says that the currency still has decimals

If ISO-4217 says the currency has decimals, it should stay this way in the `currencies_iso.yml` file. If someone wants a different (non-standard) subunit definition it should be in a converter.

Alternatively, this gem needs to stop claiming that the default built-in converter follows the standard ISO 4217 subunit definitions, and offer a prominent warning that that’s no longer going to be the case.

PR #389, opened 2025-04-07
Add Ruby 3.4

bslobodin:
Adds support for Ruby 3.4 going forward.


lib/money/splitter.rb
```diff
@@ -77,7 +77,6 @@ def [](index)
           return money
         end
       end
-      nil
```
bslobodin, line 80: image

elfassy: Screenshot 2025-04-11 at 9 04 56 AM

CI runs simplecov as and checks that we have 100% coverage, strange that it didn't pick up on it 🤔 Are you sure you ran all the tests?

elfassy: yes I think we're missing a test for the case where this returns nil. Strange that this didn't make CI fail before ruby 3.4

elfassy (approved):

lib/money/splitter.rb
elfassy: `the last expression would be split.each do ... end. The each method on a Hash returns the hash itself`

PR #398, opened 2025-05-11
Add USDC stablecoin to currency_non_iso.yml

davidpatters0n:
Reopening this PR as part of: https://vault.shopify.io/gsd/proposals/ElTwQM


antw (approved):

config/currency_non_iso.yml
```diff
@@ -94,3 +94,18 @@ gbx:
...
   smallest_denomination: 1
+usdc:
+  priority: 100
+  iso_code: USDC
+  name: USD Coin
+  symbol: USDC
+  disambiguate_symbol: USDC
+  alternate_symbols: []
+  subunit: Cent
+  subunit_to_unit: 100
```
nataliebettenburg, line 105: I saw some discussion on 100 vs 1000000 on the [linked PR](https://github.com/Shopify/money/pull/350#discussion_r1863767599), but would still like to understand this decision better.
 - why was the [number of decimal places changed from 2 to 6 in the RubyMoney gem](https://github.com/RubyMoney/money/pull/1089) ?
 - why is it ok for this config to differ from this and continue to use 2 decimal places?
      - now that we're planning to actually use this currency, is the fact that 2 decimal places behaves the same as a `MissingCurrency` still a valid argument?

mce-shop: sorry for hijacking:
- the answer for the first question: although it's not well documented, I think it's because USDC and many other stable coins use 6 decimals. I am guessing we are following the standard implementation [reference](https://www.coinbase.com/en-gb/blog/introduction-to-building-on-defi-with-ethereum-and-usdc-part-1) other non-stable coins typically use 18 decimals

davidpatters0n: In this case I may set subunits back to 6 in that case.

elfassy:
@davidpatters0n can you add to the description and explain why you want this crypto currency as part of the money gem 🙏 ?

config/currency_non_iso.yml
elfassy: @nataliebettenburg

> why was the number of decimal places changed from 2 to 6 in the RubyMoney gem

I'm going to guess it's because the token contract has 6 decimals (see https://explorer.solana.com/address/4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU?cluster=devnet). That being said, it's always displayed with 2 decimals to users (see https://www.usdc.com/). So while technically possible to send sub-cent transactions, realistically not many people would do so because all prices they show are using two decimals


> why is it ok for Shopify's config to differ from RubyMoney gem

those are two different gems, many things differ between the two

lacoursieresimon: If we don't arrive to an alignment with 6 vs 2 decimal places, we can always do 2 by default but add the `(format: stripe)` override for 6 subunits support like we do with ISK / UGX.

https://github.com/Shopify/money/blob/4377608dd60f42d58fc84e4f0d329ad65da6d697/lib/money/helpers.rb#L12-L15

```ruby
Money.new(0.50, "USDC").subunits
# => 50
Money.new(0.50, "USDC").subunits(format: :stripe)
# => 50000
```

lalib (approved):
C7/M4 Review - agree with LOW RISK. ✅

config/currency_non_iso.yml
davidpatters0n: ah good flag! FWIW I'm closing this PR in favour of: https://github.com/Shopify/money/pull/401 excuse the noise.

PR #401, opened 2025-05-14
feat: Add crypto currency support with global toggle

davidpatters0n:
Introduces a cleaner way to handle cryptocurrency support through a global configuration system, making it easier to use crypto currencies with proper thread safety and cache management.

### What Changed
- Added enable_crypto_currencies global config option
- Created with_config helper for temporary settings
- Implemented dual mutex system for thread-safe currency caching
- Moved crypto currencies to dedicated crypto_currencies.yml
- Added comprehensive test coverage for thread safety and edge cases

#### Before:

```ruby
# No built-in crypto support
Money.new(100, "USDC") # Raises UnknownCurrency

# Had to manually manage currency definitions
currencies = Money::Currency.currencies
currencies.merge!("USDC" => { ... })
```

#### After

```ruby
# Global enable
Money.config.enable_crypto_currencies = true
Money.new(100, "USDC") # Just works!

# Or use temporary blocks
Money.with_config(enable_crypto_currencies: true) do
  Money.new(100, "USDC")
end
```

#### Before

```ruby
# No thread-safe configuration
@@mutex = Mutex.new
@@loaded_currencies = {}

# Single mutex for all operations
@@mutex.synchronize { @@loaded_currencies[iso] = new(iso) }
```

#### After

```ruby
# Dual mutex system for better concurrency
@@currencies_mutex = Mutex.new
@@loaded_currencies_mutex = Mutex.new

# Separate locks for different operations
Money.with_config(enable_crypto_currencies: true) do
  # Thread-safe currency operations
  Money.new(100, "USDC")
end
```

### Example usage

```ruby
# Enable globally in initializer
Money.config.enable_crypto_currencies = true

# Use temporary blocks
Money.with_config(enable_crypto_currencies: true) do
  usdc = Money.new(100, "USDC")
  usd = Money.new(100, "USD")
  
  # Normal money operations work
  total = usdc + usd # Raises IncompatibleCurrencyError
end

# Thread-safe configuration
Thread.new do
  Money.with_config(enable_crypto_currencies: true) do
    # This thread's config is isolated
    Money.new(100, "USDC")
  end
end

# Combine settings
Money.with_config(
  enable_crypto_currencies: true,
  default_currency: "USDC"
) do
  # Multiple settings in one block
end
```


lib/money/config.rb
```diff
@@ -16,18 +16,45 @@ def legacy_json_format!
...
-      @legacy_deprecations = old_legacy_deprecations
+      original_values.each do |key, value|
+        public_send("#{key}=", value)
+      end
+    end
+
+    def enable_crypto_currencies=(value)
+      if @enable_crypto_currencies != value
+        @enable_crypto_currencies = value
+        Money::Currency.clear_cache
```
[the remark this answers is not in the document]
elfassy: is there another way to do this? For instance by searching through both sets independantly?

lib/money/currency/loader.rb
```diff
@@ -6,14 +6,24 @@ class Money
...
+          @@mutex.synchronize do
+            currencies = {}
+            currencies.merge!(YAML.load_file("#{currency_data_path}/currency_historic.yml"))
+            currencies.merge!(YAML.load_file("#{currency_data_path}/currency_non_iso.yml"))
+            currencies.merge!(YAML.load_file("#{currency_data_path}/currency_iso.yml"))
+
+            # Use global config directly
+            if Money.config.enable_crypto_currencies
+              currencies.merge!(YAML.load_file("#{currency_data_path}/crypto_currencies.yml"))
+            end
```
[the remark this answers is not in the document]
davidpatters0n: I'm trying to understand the need for maintaining two hashes, what are we gaining here? Also the suggestion above seems to always load crypto currencies ?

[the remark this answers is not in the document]
elfassy: we'd load it when the app boots, but then use it only depending on the configuration. My worry is that we get into a race condition and the global config not being set yet

lib/money/currency.rb
```diff
@@ -24,7 +26,18 @@ def find(currency_iso)
       end
 
       def currencies
-        @@currencies ||= Loader.load_currencies
+        @@currencies_mutex.synchronize do
+          @@currencies ||= Loader.load_currencies
+        end
+      end
+
+      def clear_cache
```
elfassy, line 34: let's get rid of this method, we don't want to reset this list unless the app is restarted

spec/currency_spec.rb
```diff
@@ -37,6 +37,66 @@
...
+          "decimal_mark" => "."
+        } 
+      }
+      allow(Money::Currency).to receive(:crypto_currencies).and_return(crypto_currencies)
+      
+      currency = Money::Currency.new('USDC')
+      expect(currency.iso_code).to eq('USDC')
+      expect(currency.symbol).to eq('USDC')
+      
+      Money.config = old_config
```
antw, line 65: question: Does this been to be in an `ensure` block?

If the test fails I'm assuming the config won't be restored to the previous state and the custom config might affect other tests. I'm guessing we need something like:

```ruby
old_config = Money.config

begin
  # the test
ensure
  Money.config = old_config
end
```

I wonder if it's worth adding a `with_config` method [here](https://github.com/Shopify/money/blob/4377608dd60f42d58fc84e4f0d329ad65da6d697/lib/money/money.rb#L52)?

```ruby
def with_config(config)
  old_config = Money.config
  Money.config = config
  yield
ensure
  Money.config = old_config
end
```

spec/spec_helper.rb
```diff
@@ -72,13 +72,14 @@ def missing_methods
...
 
-def configure(default_currency: nil, legacy_json_format: nil, legacy_deprecations: nil, legacy_default_currency: nil)
+def configure(default_currency: nil, legacy_json_format: nil, legacy_deprecations: nil, legacy_default_currency: nil, crypto_currencies: nil)
   old_config = Money.config
   Money.config = Money::Config.new.tap do |config|
     config.default_currency = default_currency if default_currency
     config.legacy_json_format! if legacy_json_format
     config.legacy_deprecations! if legacy_deprecations
     config.legacy_default_currency! if legacy_default_currency
+    config.crypto_currencies = true if crypto_currencies
```
antw, line 82: nit:

```suggestion
    config.crypto_currencies = crypto_currencies unless crypto_currencies.nil?
```

robinbrandt: Shouldn't this be a boolean and then just `config.crypto_currencies = crypto_currencies`?

lib/money/currency.rb
```diff
@@ -26,6 +26,10 @@ def find(currency_iso)
       def currencies
         @@currencies ||= Loader.load_currencies
       end
+
+      def crypto_currencies
```
robinbrandt, line 30: I didn't see in the PR description if this was a deliberate choice. Another alternative would've been to extend the loader to be aware of what to load and just have one internal hash of all currencies (that is the union of fiat and crypto currencies) which would mean that other dependent code that's using `Money::Currency.currencies` would just work. A typical example for this is code that generates a schema and generates an enum of currencies.

One can argue either way but just curious. 

Regardless of this I'd document the behaviour in the new crypto section in the doc. It might be surprising for consumers of the API.

spec/currency_spec.rb
```diff
@@ -37,6 +37,65 @@
...
+          "decimal_mark" => "."
+        } 
+      }
+      allow(Money::Currency).to receive(:crypto_currencies).and_return(crypto_currencies)
+      
+      # When crypto_currencies is disabled, new should raise an error for a crypto currency
+      expect { Money::Currency.new('USDC') }.to raise_error(Money::Currency::UnknownCurrency)
+      
+      # Cleanup
+      Money.config = old_config
```
antw, line 96: question: Can/should we use `with_config` here? Same in the test below.

```suggestion
      config = Money::Config.new
      config.experimental_crypto_currencies = false

      with_config(config) do
        # Mock currencies
        allow(Money::Currency).to receive(:currencies).and_return({})
        crypto_currencies = { 
          "usdc" => { 
            "iso_code" => "USDC", 
            "name" => "USD Coin",
            "symbol" => "USDC",
            "disambiguate_symbol" => "USDC",
            "subunit_to_unit" => 100,
            "smallest_denomination" => 1,
            "decimal_mark" => "."
          } 
        }
        allow(Money::Currency).to receive(:crypto_currencies).and_return(crypto_currencies)

        # When crypto_currencies is disabled, new should raise an error for a crypto currency
        expect { Money::Currency.new('USDC') }.to raise_error(Money::Currency::UnknownCurrency)
      end

```

spec/currency_spec.rb
```diff
@@ -37,6 +37,65 @@
...
+        currency = Money::Currency.new('USDC')
+        expect(currency.iso_code).to eq('USDC')
+        expect(currency.symbol).to eq('USDC')
+      end
+    end
+    
+    it "doesn't look up crypto currencies when disabled" do
+      # Reset class variables to ensure clean state
+      old_currencies = Money::Currency.class_variable_get(:@@loaded_currencies) rescue {}
+      Money::Currency.class_variable_set(:@@loaded_currencies, {})
```
antw, line 70: question: Is it worth adding this (and the clean-up at the end of the test) to `with_config`?

elfassy (approved):

spec/currency_spec.rb
elfassy: I agree, this test should be improved, it looks like it was generated by AI. The test can be as simple as
```
configure(experimental_crypto_currencies: false) do 
  assert_nil Money::Currency.find("USDC") 
end
```

davidpatters0n: @elfassy - This test wasn't generated by AI, the need for doing the above is because `Money::Currency` has the class variable `@@loaded_currencies` to cache the currencies therefore we need to clear. But yes, we moved the above into the configure method and the test has now become the above.

PR #417, opened 2025-06-11
Add missing `require`

amomchilov:
Fixes crash when used in a context where `bigdecimal` wasn't already required.

```
Error: bundler: failed to load command: tapioca (/Users/alex/.gem/ruby/3.4.3/bin/tapioca)
/Users/alex/.gem/ruby/3.4.3/gems/shopify-money-3.2.1/lib/money/allocator.rb:11:in '<class:Allocator>': undefined method 'BigDecimal' for class Money::Allocator (NoMethodError)

    ONE = BigDecimal("1")
          ^^^^^^^^^^
	from /Users/alex/.gem/ruby/3.4.3/gems/shopify-money-3.2.1/lib/money/allocator.rb:6:in '<class:Money>'
```

### DIFF
--- a/lib/money/version.rb
+++ b/lib/money/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 class Money
-  VERSION = "3.2.1"
+  VERSION = "3.2.2"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    shopify-money (3.2.1)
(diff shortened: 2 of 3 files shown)


lib/money/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 class Money
-  VERSION = "3.2.1"
+  VERSION = "3.2.2"
```
amomchilov, line 4: Gotta bump the shame version

SemVer shame version

Issue #421, opened 2025-06-16
nil pointer when using a write accessor with `currency_read_only`


lavoiesl:
Within `spec/money_column_spec.rb`:
```rb
it 'sets the price correctly when the currency is not changed' do
  record = MoneyWithReadOnlyCurrency.create(currency: 'CAD', price: 1)
  record.price = Money.new(2, 'CAD')
  expect(record.price.value).to eq(2)
end
```

```rb
NoMethodError:
 undefined method '[]' for nil
# ./lib/money_column/active_record_hooks.rb:78:in 'MoneyColumn::ActiveRecordHooks#compatible_currency?'
# ./lib/money_column/active_record_hooks.rb:60:in 'MoneyColumn::ActiveRecordHooks#write_money_attribute'
# ./lib/money_column/active_record_hooks.rb:120:in 'block (2 levels) in MoneyWithReadOnlyCurrency#money_column'
```

PR #422, opened 2025-06-16
[bugfix] money_column allow assigning money

elfassy:
# Why
fixes https://github.com/Shopify/money/issues/421

# What
when a money is assigned to a new instance `_assign_attributes` is yet called. This PR correctly reverts back to the old behaviour when this is the case


lib/money_column/active_record_hooks.rb
```diff
@@ -74,9 +74,9 @@ def write_money_attribute(column, money)
 
     def compatible_currency?(money, options)
       currency_column = options[:currency_column]
-      currency = options[:currency] ||
-        @money_raw_new_attributes[currency_column.to_sym] ||
-        try(currency_column)
+      currency = options[:currency]
+      currency ||= @money_raw_new_attributes[currency_column.to_sym] if @money_raw_new_attributes
+      currency ||= try(currency_column)
```
lavoiesl, line 79: Why is `@money_raw_new_attributes` even needed? 🤔

elfassy: it's required to make https://github.com/Shopify/money/pull/409/files#diff-0fdde4bd8a1f5fb142f661be86c568d43b1061f447e188d1d585ab1974458548R416 test pass. Let me know if you have another way you're thinking of that doesn't requires this, but also keeps the test passing

lavoiesl: I'm cool with merging this for now and exploring other options in parallel 👍🏼

PR #500, opened 2026-01-16
Fix and clarify README

rono23:


README.md
```diff
@@ -212,13 +212,12 @@ ActiveRecord:
 
 ```ruby
 create_table :orders do |t|
-  t.decimal :sub_total, precision: 21, scale: 3
-  t.decimal :tax, precision: 21, scale: 3
-  t.string :currency, limit: 3
+  t.decimal :price, precision: 21, scale: 3
+  t.string :price_currency, limit: 3
```
rono23: These names are already used in [the test code](https://github.com/Shopify/money/blob/68bf1b027d65793a531a4baf9c1b99ba2a54dc60/spec/money_column_spec.rb), so I aligned the README for better clarity.

PR #519, opened 2026-03-04
perf: skip redundant currency resolution in Money.new

cribbles:
## Summary

`Money.new(value, "USD")` resolves the currency string twice: once in `Money.new` via `Helpers.value_to_currency`, and again in `Money#initialize` which calls `value_to_currency` on the already-resolved `Currency` object. This means every construction pays for two full `case`/`when` dispatches through `value_to_currency` and two `Currency.find!` lookups instead of one.

This PR makes two changes:

1. **Remove redundant `value_to_currency` from `Money#initialize`**: since `Money.new` always resolves the currency before calling `super`, `initialize` can trust that its `currency` argument is already a `Currency` object. `init_with` (YAML deserialization) is updated to resolve before calling `initialize`.

2. **Cache string-to-Currency mappings in `Helpers.value_to_currency`**: repeat calls with the same currency code string (e.g. `"USD"`) now skip the `Currency.find!` lookup (which does `to_s.downcase` + hash lookup) after the first resolution. The cache is reset alongside `Currency.reset_loaded_currencies`.

### Benchmark (500k iterations of `Money.new`)

```
                        Before      After       Improvement
string ("USD")          ~455ms      ~360ms      ~21%
Currency object         ~325ms      ~285ms      ~12%
string/currency gap     ~130ms      ~75ms       ~42% smaller
```

### DIFF
--- a/lib/money/helpers.rb
+++ b/lib/money/helpers.rb
@@ -44,10 +44,20 @@ class Money
       when 'xxx', 'XXX'
         Money::NULL_CURRENCY
       when String
-        Currency.find!(currency)
+        resolved_currencies[currency] ||= Currency.find!(currency)
       else
         raise ArgumentError, "could not parse as currency #{currency.inspect}"
       end
+    end
+
+    def reset_resolved_currencies
+      @resolved_currencies = {}
+    end
+
+    private
+
+    def resolved_currencies
+      @resolved_currencies ||= {}
     end
   end
 end
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -128,13 +128,16 @@ class Money
     raise ArgumentError if value.nan?
     raise ArgumentError if value.infinite?
 
-    @currency = Helpers.value_to_currency(currency)
+    @currency = currency
     @value = BigDecimal(value.round(@currency.minor_units))
     freeze
   end
 
   def init_with(coder)
-    initialize(Helpers.value_to_decimal(coder['value']), coder['currency'])
+    initialize(
+      Helpers.value_to_decimal(coder['value']),
+      Helpers.value_to_currency(coder['currency']),
+    )
   end
 
   def encode_with(coder)
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -42,6 +42,7 @@ class Money
 
       def reset_loaded_currencies
         @@loaded_currencies = {}
+        Helpers.reset_resolved_currencies
       end
     end
 


lib/money/helpers.rb
```diff
@@ -44,10 +44,20 @@ def value_to_currency(currency)
       when 'xxx', 'XXX'
         Money::NULL_CURRENCY
       when String
-        Currency.find!(currency)
+        resolved_currencies[currency] ||= Currency.find!(currency)
```
elfassy, line 47: this is already happening inside the `find!` method https://github.com/Shopify/money/blob/main/lib/money/currency.rb#L16 🤔 wonder how you're still getting a speed improvement. Can you remove the `resolved_currencies` and redo the benchmark 🙏

lib/money/money.rb
```diff
@@ -128,13 +128,16 @@ def initialize(value, currency)
     raise ArgumentError if value.nan?
     raise ArgumentError if value.infinite?
 
-    @currency = Helpers.value_to_currency(currency)
+    @currency = currency
```
elfassy, line 131: 👍

lib/money/money.rb
```diff
@@ -128,13 +128,16 @@ def initialize(value, currency)
...
     @value = BigDecimal(value.round(@currency.minor_units))
```
elfassy, line 132: we can probably also do the same here

```suggestion
    @value = value.round(@currency.minor_units)
```

cribbles: Tried this out... nice idea; makes the tests fail. The problem is `BigDecimal#round` returns an `Integer` when the currency has no subunits (i.e. `minor_units == 0` -- JPY for example), so we really need this wrapper for those cases. (Methods downstream assume `@value` will be a `BigDecimal` instance)

lib/money/helpers.rb
cribbles: In those cases we skip over an [invocation](https://github.com/Shopify/money/blob/main/lib/money/currency.rb#L15C15-L15C41) of `currency_iso.to_s.downcase`, which helps at the margins.

cribbles:
@elfassy 

Benchmark I used for this:

```rb
require "money"
require "benchmark"

n = 500_000
c = Money::Currency.find!("USD")

Benchmark.bm do |x|
  x.report("string") { n.times { Money.new(100, "USD") } }
  x.report("currency") { n.times { Money.new(100, c) } }
end
```

---

Results when I ran them just now:

#### Main branch

```sh
string    0.426026   0.000652   0.426678 (  0.426774)
currency  0.303333   0.000438   0.303771 (  0.303835)
```

#### This branch

```sh
string    0.349412   0.001080   0.350492 (  0.350653)
currency  0.276891   0.000863   0.277754 (  0.277763)
```

lib/money/helpers.rb
elfassy: if we want to skip the downcase, we should do it in the `find!`. This could effectively double the size of the hash, but could be worthwhile. Can you split the PRs (one for removing the downcase, and one for the initialize changes). We need to show that each is a performance improvement on their own

```ruby
      def new(currency_iso)
        iso = currency_iso.to_s
        return @@loaded_currencies[iso] if @@loaded_currencies.key?(iso)

        raise UnknownCurrency, "Currency can't be blank" if iso.empty?
        @@mutex.synchronize { @@loaded_currencies[iso] = super(iso.downcase) }
      end
```

PR #522, opened 2026-04-21
Fix float imprecision in allocate and add configurable strategy

elfassy:
# Why

Bug
```ruby
Money::Allocator.new(Money.new(100, "USD")).allocate([0.02, 0.98])
=> [#<Money value:2.01 currency:USD>, #<Money value:97.99 currency:USD>]
```

After
```ruby
Money::Allocator.new(Money.new(100, "USD")).allocate([0.02, 0.98])
=> [#<Money value:2.00 currency:USD>, #<Money value:98.00 currency:USD>]
```

# What

```
> 0.98.to_r == Rational(98, 100)
=> false
```
this is because floats have errors that get carried into the rational. 
```
0.98.to_r  #=> (2206763817411543/2251799813685248)  # slightly less than 0.98!
```

Option1: `0.98.to_s.to_r #=> (98/100)`
Option2: `0.98.rationalize #=> (98/100)`

- fix the bug, went with option 2 which gets us better performance
- Allow setting the strategy at the app level: `config.default_allocation_strategy = :roundrobin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,23 @@ Money.new(money.value * exchange_rate, "JPY")
 money.convert_currency(exchange_rate, "JPY")
 ```
 
+### Allocation Strategy
+
+By default `allocate` distributes leftover subunits using the `:roundrobin` strategy.
+You can change the default:
+
+``` ruby
+Money.configure do |config|
+  config.default_allocation_strategy = :nearest
+end
+```
+
+Or pass a strategy per call:
+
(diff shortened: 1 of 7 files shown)


paupino (approved):

robinbrandt (approved):

elfassy:
cc @eargollo thanks for flagging this bug, got to the bottom of it

Issue #437, opened 2025-08-06
Is it okay to add view helpers?


rupokghosh:
We are going to be using this in our repo and I was wondering if there are ways to format Objects to view helpers so we can get formatted values like:
1. `$125, 000`
2. `$125, 000 USD`
3. `USD 125, 000`
4.  `125, 000 USD`
etc, etc

If not, is it okay if I spin up a couple of PRs to add a way to format Objects to one of these, would help a lot!

elfassy:
The formatting of currency strings is locale specific. As such the preferred approach is to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat to do the rendering.

If for whatever reason you must really do this on the backend, Rails has a method for you https://api.rubyonrails.org/classes/ActiveSupport/NumberHelper.html#method-i-number_to_currency

PR #526, opened 2026-05-27
Expose Currency#thousands_separator

pedropb:
## Why

Currency data already includes `thousands_separator`, but `Money::Currency` only exposed `decimal_mark`. Callers that need currency-aware formatting currently have to reach back into `Money::Currency.currencies` even though the loaded currency object already represents the same data.

## What changed

- Expose `Money::Currency#thousands_separator` from the loaded currency data.
- Keep `Money::NullCurrency` quacking like `Money::Currency` by exposing the same reader with the USD-style delimiter.
- Update RBS signatures, docs, and currency specs.

## Formatting note

I also checked for a higher-level formatter. The gem currently exposes `Money#to_s` / `Money#to_fs`, but those return the numeric amount string only. They do not include currency symbols, thousands delimiters, or Rails-style `negative_format` handling. This keeps Rails callers able to use `number_to_currency` while sourcing `separator` and `delimiter` directly from `money.currency`.

## Verification

Covered by specs for built-in currency attributes, custom currency loading, and NullCurrency interface parity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ currency.name     #=> "United States Dollar"
 currency.to_s     #=> 'USD'
 currency.symbol   #=> '$'
 currency.disambiguate_symbol #=> 'US$'
+currency.decimal_mark #=> '.'
+currency.thousands_separator #=> ','
 ```
 
 ### Default Currency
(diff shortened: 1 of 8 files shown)


elfassy:
How you format your currency is locale specific, not country/currency specific. In other words, `1000 CAD` can be `1,000.00 CAD` or `1.000,00 CAD` depending where the person looking at this value is from (in this case `en-CA` vs `fr-CA`).  Have a look at this answer https://github.com/Shopify/money/issues/437#issuecomment-4537344156

elfassy:
closing for now, will add more info about formatting to the readme in another PR

PR #528, opened 2026-07-28
perf: exact-key currency cache, inline lookup in Money.new

alexcwatt:
### What

Speeds up `Currency.find!`/`Currency.new` with an exact-key cache, so repeat lookups of the same spelling (e.g. `"USD"`, `:usd`) are a single Hash read with zero allocations — the cache previously called `to_s.downcase` on **every** lookup, allocating even on cache hits.

This is a revised version of the approach in #521, incorporating the review feedback there and on #519:

- **The optimization lives in `find!` itself** (not another caching layer): `Currency.new`/`find!` checks the cache under the exact key first, and only downcases on a miss.
- **Cache growth is bounded**: only symbols and the canonical UPCASE string are memoized under their exact key (the downcased string is already the primary key). Mixed-case spellings like `"UsD"` still resolve correctly but are never memoized, so the cache holds at most a few entries per currency (~194 × 3 realistic worst case). `reset_loaded_currencies` still clears everything.
- `Helpers.value_to_currency` — the designated coercion point used by `Money.new` — consults the same cache first for String arguments before falling into its `case`. The cache never contains `''`, `'xxx'`, or `'XXX'` (`Currency.new` raises before memoizing), so all special-string handling is preserved. No caching logic leaks into `Money.new`.
- Class variables → a `LOADED_CURRENCIES` constant (mutated in place, never reassigned) and a singleton-class ivar mutex, both faster to access than `@@` variables.

### Performance

Per the feedback on #521, all benchmarks below are measured **with the cache fully populated with every currency under three spellings** (`"USD"`, `"usd"`, `:USD` → 582 entries), the worst case for cache size. Hash lookups are O(1), so lookups get *faster*, not slower, despite the larger hash. Ruby 4.0.1 (arm64-darwin), median of 7 trials:

| `Currency.find!` (582-entry cache) | main | #521 | this PR |
|---|---|---|---|
| `find! "USD"` | 106.9 ns / 1 alloc | 62.4 ns / 0 alloc | **35.5 ns / 0 alloc** |
| `find! "usd"` | 106.3 ns / 1 alloc | 63.8 ns / 0 alloc | **35.4 ns / 0 alloc** |
| `find! :USD` | 144.5 ns / 3 allocs | 104.2 ns / 1 alloc | **34.3 ns / 0 alloc** |
| all 582 spellings, cycled | 183.6 ns / 2 allocs | 104.9 ns / 0.33 alloc | **58.0 ns / 0 alloc** |

| `Money.new` mixed workload¹ | main | #521 | this PR |
|---|---|---|---|
| wall time (median of 5 runs) | 822 ns/op | 771 ns/op | **649 ns/op (-21% vs main)** |
| allocations | 3.80 /op | 2.93 /op | **2.93 /op (-23% vs main)** |

¹ integers/floats/strings/BigDecimals/Rationals × ISO strings/Currency/nil, matching real-world argument mixes.

Breakdown: the exact-key cache in `find!` provides the `find!` speedup; the `value_to_currency` fast path provides most of the `Money.new` win (it skips the case/when string comparisons for the dominant String-argument path); the class-var → constant/ivar change is a small constant win on every access.

### Memory (addressing the #521 review concern)

Measured retained size of the cache hash + string keys, fully populated:

| | entries | retained |
|---|---|---|
| main | 194 | 14.7 KB |
| #521 | 388 | 29.3 KB |
| this PR (worst case: every currency looked up as `"usd"`, `"USD"`, and `:usd`) | 582 | 43.3 KB |

Bounded by spellings actually used, and hard-capped by the memoization guard: arbitrary mixed-case strings are never added to the cache.

### Correctness

- Full spec suite (731 examples), rubocop, and steep green; 100% line coverage maintained. New specs cover cached-instance identity across spellings and the bounded memoization.
- A 281-case differential test vs main (all value types × currency args incl. `""`/`nil`/`"xxx"`/`"XXX"`/mixed-case/symbols/bignums/`-0.0`/Money passthrough, plus `find`/`find!`) produced identical results.
- `Money.new(1, :usd)` still raises `ArgumentError` (the helper fast path is String-guarded), and `value_to_currency("XXX")` still returns `NULL_CURRENCY`.

<details>
<summary>Benchmark script (run <code>ruby -Ilib bench.rb</code> on each ref)</summary>

```ruby
# frozen_string_literal: true

# Populates the currency cache with EVERY currency under three spellings
# ("USD", "usd", :USD) -- the worst case for cache size, per review feedback
# on #521 -- then measures lookups against that fully-loaded hash and reports
# the actual retained memory of the cache.

require "money"
require "bigdecimal"
require "objspace"

Money.configure { |c| c.default_currency = "USD" }

iso_codes = Money::Currency.currencies.keys
spellings = []
iso_codes.each do |code|
  spellings << code.upcase
  spellings << code
  spellings << code.upcase.to_sym
end
3.times { spellings.each { |s| Money::Currency.find!(s) } }

cache =
  begin
    Money::Currency::LOADED_CURRENCIES
  rescue NameError
    Money::Currency.class_variable_get(:@@loaded_currencies)
  end

cache_bytes = ObjectSpace.memsize_of(cache) +
  cache.keys.sum { |k| k.is_a?(String) ? ObjectSpace.memsize_of(k) : 0 }
puts format("cache entries: %d, hash+keys retained: %.1f KB", cache.size, cache_bytes / 1024.0)

def median_ns_per_op(total_ops, trials: 7)
  times = trials.times.map do
    GC.start
    t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    yield
    Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
  end
  (times.sort[trials / 2] / total_ops) * 1_000_000_000
end

def allocs_per_op(total_ops)
  GC.start
  GC.disable
  before = GC.stat(:total_allocated_objects)
  yield
  allocs = GC.stat(:total_allocated_objects) - before
  GC.enable
  allocs.to_f / total_ops
end

def report(name, n, bench)
  ns = median_ns_per_op(n) { bench.call }
  al = allocs_per_op(n) { bench.call }
  puts format("  %-34s %8.1f ns/op  %6.3f allocs/op", name, ns, al)
end

N = 2_000_000

puts "-- Currency.find! (cache fully populated, #{cache.size} entries) --"
report("find! 'USD' (upcase string)", N, -> { i = 0; while i < N; Money::Currency.find!("USD"); i += 1; end })
report("find! 'usd' (downcase string)", N, -> { i = 0; while i < N; Money::Currency.find!("usd"); i += 1; end })
report("find! :USD (symbol)", N, -> { i = 0; while i < N; Money::Currency.find!(:USD); i += 1; end })

ALL = spellings.freeze
CYCLES = N / ALL.size
report("find! all #{ALL.size} spellings (cycled)", CYCLES * ALL.size, -> {
  i = 0
  while i < CYCLES
    ALL.each { |s| Money::Currency.find!(s) }
    i += 1
  end
})

usd = Money::Currency.find!("USD")
big = BigDecimal("12.34")
existing = Money.new(5, "USD")

INPUTS = [
  [1, "USD"], [100, "USD"], [42, "CAD"], [12.34, "USD"], [99.99, "EUR"],
  ["12.34", "USD"], ["0.99", "CAD"], [big, "USD"], [big, usd], [1500, "JPY"],
  [0, "USD"], [existing, "USD"], [7, nil], [Rational(1, 3), "USD"], ["5", "GBP"],
].freeze

MIX_N = 20_000
puts "-- Money.new (mixed inputs) --"
report("Money.new mixed workload", MIX_N * INPUTS.size, -> {
  i = 0
  while i < MIX_N
    INPUTS.each { |v, c| Money.new(v, c) }
    i += 1
  end
})
```

</details>

### DIFF
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -18,8 +18,12 @@ class Money
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.to_s.empty?
         iso = currency_iso.to_s.downcase
         currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
-        # Memoize the original spelling so the next lookup takes the fast path.
-        if currency_iso.is_a?(String) || currency_iso.is_a?(Symbol)
+        # Memoize common alternate spellings (:usd, "USD") under their exact key
+        # so the next lookup takes the fast path above without allocating.
+        # Bounded: only symbols and the canonical UPCASE string are memoized
+        # (the downcased string is already the primary key), so the cache holds
+        # at most a few entries per currency.
+        if currency_iso.is_a?(Symbol) || currency_iso == currency.iso_code
           @mutex.synchronize { LOADED_CURRENCIES[currency_iso] = currency }
         end
         currency
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -18,11 +18,8 @@ class Money
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.to_s.empty?
         iso = currency_iso.to_s.downcase
         currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
-        # Memoize common alternate spellings (:usd, "USD") under their exact key
-        # so the next lookup takes the fast path above without allocating.
-        # Bounded: only symbols and the canonical UPCASE string are memoized
-        # (the downcased string is already the primary key), so the cache holds
-        # at most a few entries per currency.
+        # Only canonical spellings (:usd, "USD") are memoized, so arbitrary-case
+        # strings can't grow the cache without bound.
         if currency_iso.is_a?(Symbol) || currency_iso == currency.iso_code
           @mutex.synchronize { LOADED_CURRENCIES[currency_iso] = currency }
         end
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -6,23 +6,19 @@ class Money
   class Currency
     class UnknownCurrency < ArgumentError; end
 
-    # Cache of Currency instances, keyed by ISO code spelling (String/Symbol).
-    # Mutated in place (never reassigned) so hot paths can reference it directly.
     LOADED_CURRENCIES = {}
 
     @mutex = Mutex.new
 
     class << self
       def new(currency_iso)
-        # Fast path: exact-key cache hit (e.g. "USD") avoids to_s/downcase allocations.
         cached = LOADED_CURRENCIES[currency_iso]
         return cached if cached
 
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.to_s.empty?
         iso = currency_iso.to_s.downcase
         currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
-        # Memoize the original spelling (Hash dups+freezes string keys) so the
-        # next lookup with the same spelling takes the fast path.
+        # Memoize the original spelling so the next lookup takes the fast path.
         if currency_iso.is_a?(String) || currency_iso.is_a?(Symbol)
           @mutex.synchronize { LOADED_CURRENCIES[currency_iso] = currency }
         end
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -4,16 +4,29 @@ require "money/currency/loader"
 
 class Money
   class Currency
-    @@mutex = Mutex.new
-    @@loaded_currencies = {}
-
     class UnknownCurrency < ArgumentError; end
+
+    # Cache of Currency instances, keyed by ISO code spelling (String/Symbol).
+    # Mutated in place (never reassigned) so hot paths can reference it directly.
+    LOADED_CURRENCIES = {}
+
+    @mutex = Mutex.new
 
     class << self
       def new(currency_iso)
+        # Fast path: exact-key cache hit (e.g. "USD") avoids to_s/downcase allocations.
+        cached = LOADED_CURRENCIES[currency_iso]
+        return cached if cached
+
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.to_s.empty?
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
+        # Memoize the original spelling (Hash dups+freezes string keys) so the
+        # next lookup with the same spelling takes the fast path.
+        if currency_iso.is_a?(String) || currency_iso.is_a?(Symbol)
+          @mutex.synchronize { LOADED_CURRENCIES[currency_iso] = currency }
+        end
+        currency
       end
       alias_method :find!, :new
 
@@ -41,7 +54,7 @@ class Money
       end
 
       def reset_loaded_currencies
-        @@loaded_currencies = {}
+        LOADED_CURRENCIES.clear
       end
     end
 
--- a/lib/money/currency.rb
+++ b/lib/money/currency.rb
@@ -4,27 +4,28 @@ require "money/currency/loader"
 
 class Money
   class Currency
+    @@mutex = Mutex.new
+    @@loaded_currencies = {}
+
     class UnknownCurrency < ArgumentError; end
-
-    LOADED_CURRENCIES = {}
-
-    @mutex = Mutex.new
 
     class << self
       def new(currency_iso)
-        cached = LOADED_CURRENCIES[currency_iso]
-        return cached if cached
+        currency = @@loaded_currencies[currency_iso]
+        return currency if currency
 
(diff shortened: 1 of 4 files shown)


lib/money/currency.rb
```diff
@@ -4,16 +4,26 @@
...
       def new(currency_iso)
+        cached = LOADED_CURRENCIES[currency_iso]
+        return cached if cached
+
         raise UnknownCurrency, "Currency can't be blank" if currency_iso.nil? || currency_iso.to_s.empty?
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
+        # Only canonical spellings (:usd, "USD") are memoized, so arbitrary-case
+        # strings can't grow the cache without bound.
```
elfassy: I'm not sure I follow, `iso` is always a lowercase string because of line 19 `iso = currency_iso.to_s.downcase`

elfassy: oh i see you're doing an early return if it's already lowercase

lib/money/currency.rb
```diff
@@ -4,16 +4,26 @@
...
+        if currency_iso.is_a?(Symbol) || currency_iso == currency.iso_code
```
elfassy: I might be missing something, but I don't think this will ever happen. a valid symbol will be caught by the line above

lib/money/helpers.rb
```diff
@@ -34,6 +34,13 @@ def value_to_decimal(num)
     end
 
     def value_to_currency(currency)
+      if currency.is_a?(String)
```
elfassy: this can be removed, string is already handled and `find!` already cached

lib/money/currency.rb
alexcwatt: Yes - the early return at line 16 will be for (1) lowercase or (2) all uppercase or (3) symbol downcase version

lib/money/currency.rb
alexcwatt: Without this code change, we do not memoize with symbols, only with lowercase strings. (Maybe I am missing something)

I can remove the memoization on symbols - maybe we don't do it much

elfassy: > we do not memoize with symbols, only with lowercase strings

symbols get converted to strings before we get here, so the memoization will work no?

lib/money/currency.rb
```diff
@@ -4,16 +4,27 @@
...
         iso = currency_iso.to_s.downcase
-        @@loaded_currencies[iso] || @@mutex.synchronize { @@loaded_currencies[iso] = super(iso) }
+        currency = LOADED_CURRENCIES[iso] || @mutex.synchronize { LOADED_CURRENCIES[iso] = super(iso) }
+        # Memoize the canonical spelling ("USD") under its exact key so future
+        # lookups hit the early return above without allocating. Symbols and
+        # mixed-case strings ("UsD") are excluded to keep the cache bounded.
+        if currency_iso == currency.iso_code
+          @mutex.synchronize { LOADED_CURRENCIES[currency_iso] = currency }
+        end
+        currency
```
elfassy:
```suggestion
```
you can get rid of all this, it doesn't do anything. `currency` is already cached in the LOADED_CURRENCIES

lib/money/currency.rb
alexcwatt: Yes, they do, but not without object allocations (like `String#downcase`)

image

The reason I'm adding more things to the cache is to avoid allocations

lib/money/currency.rb
alexcwatt: This was the main thing I wanted to add: caching on more keys than just the downcase-string version of the currency. It causes allocations. See the benchmark script.

lib/money/currency.rb
```diff
@@ -4,16 +4,27 @@
 
 class Money
   class Currency
-    @@mutex = Mutex.new
-    @@loaded_currencies = {}
```
elfassy, line 8: there are some slight differences between a constant and a class variable, in this case we really want a class variable

lib/money/currency.rb
```diff
@@ -4,16 +4,27 @@
...
-
     class UnknownCurrency < ArgumentError; end
 
+    LOADED_CURRENCIES = {}
+
+    @mutex = Mutex.new
```
elfassy: any reason to make this a class instance variable instead of a class variable?

lib/money/currency.rb
elfassy: If you want to skip the downcase (and increase the size of the hash) you'd need to remove this line `iso = currency_iso.to_s.downcase`

alexcwatt: In its current shape, the PR allows just the upcase ("USD") and downcase ("usd") version into the cache. I think this will be sufficient for my purposes and it keeps the cache bounded.

elfassy: oh I think I get what you're trying to do now, sorry took me a while 😂 . I'll have another look tomorrow to see if there's a way to make it easier to understand

alexcwatt: Haha, no worries! Yes, please, if you have a better idea for how to make this more obvious, that's great. I do have test coverage that also shows what I'm doing here.

elfassy:
pushed a commit that will do the trick. I don't think it's worth optimizing for camelcase spelling, that can keep the downcase tax. If you think it looks good we can merge it

alexcwatt:
@elfassy thanks - yes, this is clean and works for me! Makes sense to pay the tax upfront of writing all the standard variants into the cache and makes it easier to reason about.

elfassy (approved):

PR #532, opened 2026-07-28
perf: plain defs instead of Forwardable delegators for hot value methods

alexcwatt:
### What

Replaces the `def_delegators :@value, ...` line with plain method definitions for `zero?`, `nonzero?`, `positive?`, `negative?`, `to_i`, `to_f`, and `hash`.

Forwardable-generated methods carry ~25–30ns of extra indirection per call. These are among the most frequently called `Money` methods (equality/hashing in collections, sign checks in arithmetic-heavy code), so define them directly.

Behavior is identical — each method still delegates to `@value` (`nonzero?` still returns `@value` or `nil`, matching BigDecimal semantics).

### Performance

Measured on Ruby 4.0.1 (arm64-darwin), 2M calls each, best of 3:

| method | main | this PR | Δ |
|---|---|---|---|
| `zero?` | 49.3 ns | 24.6 ns | **-50%** |
| `to_i` | 49.7 ns | 24.8 ns | **-50%** |
| `hash` | 53.6 ns | 29.1 ns | **-46%** |
| `positive?` | 130.5 ns | 98.2 ns | **-25%** |
| `negative?` | 128.5 ns | 98.3 ns | **-24%** |

(A follow-up PR takes `positive?`/`negative?` down to ~29ns via `BigDecimal#sign`.)

### Correctness

Full spec suite, rubocop, and steep green; 100% line coverage maintained.

<details>
<summary>Benchmark script (save as <code>bench.rb</code>, run <code>ruby -Ilib bench.rb</code> on each ref)</summary>

```ruby
# frozen_string_literal: true

# Standalone benchmark for shopify-money hot paths. No external dependencies.
#
# Usage (from the repo root, on each ref you want to compare):
#   ruby -Ilib money_pr_bench.rb
#
# Section 1 measures Money.new across a representative mix of input types
# (median of 7 trials, plus deterministic object allocations per call).
# Section 2 measures a few hot Money instance methods.

require "money"
require "bigdecimal"

Money.configure do |config|
  config.default_currency = "USD"
end

usd = Money::Currency.find!("USD")
big = BigDecimal("12.34")
existing = Money.new(5, "USD")

# Representative input mix (value, currency) — weighted roughly by
# real-world usage: integers/strings/bigdecimals with ISO strings dominate.
INPUTS = [
  [1, "USD"],
  [100, "USD"],
  [42, "CAD"],
  [12.34, "USD"],
  [99.99, "EUR"],
  ["12.34", "USD"],
  ["0.99", "CAD"],
  [big, "USD"],
  [big, usd],
  [1500, "JPY"],
  [0, "USD"],          # zero-cache path
  [existing, "USD"],   # Money passthrough
  [7, nil],            # default currency
  [Rational(1, 3), "USD"],
  ["5", "GBP"],
].freeze

N = 20_000 # iterations of the whole mix per trial

def run_mix
  inputs = INPUTS
  i = 0
  while i < N
    inputs.each { |v, c| Money.new(v, c) }
    i += 1
  end
end

# Warmup (also populates currency cache, zero-money cache, etc.)
2.times { run_mix }

# --- Allocations (deterministic) ---
GC.start
GC.disable
before = GC.stat(:total_allocated_objects)
run_mix
allocs = GC.stat(:total_allocated_objects) - before
GC.enable

total_ops = N * INPUTS.size
allocs_per_op = allocs.to_f / total_ops

# --- Wall time: median of 7 trials ---
times = 7.times.map do
  GC.start
  t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  run_mix
  Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
end
median = times.sort[times.size / 2]
ns_per_op = (median / total_ops) * 1_000_000_000

puts "== Money.new (mixed input types) =="
puts format("  %-18s %8.1f ns/op", "wall time:", ns_per_op)
puts format("  %-18s %8.3f objects/op", "allocations:", allocs_per_op)
puts format("  %-18s %s", "trials (ms):", times.map { |t| (t * 1000).round(1) }.inspect)

# --- Section 2: hot instance methods ---
M = 2_000_000
money = Money.new("12.34", "USD")
neg = Money.new("-12.34", "USD")

def bench_method(label, m)
  best = 3.times.map do
    GC.start
    t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    yield
    Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
  end.min
  puts format("  %-12s %8.1f ns/op", label, (best / m) * 1_000_000_000)
end

puts
puts "== Money instance methods (#{M} calls each, best of 3) =="
bench_method("zero?", M) { i = 0; while i < M; money.zero?; i += 1; end }
bench_method("to_i", M) { i = 0; while i < M; money.to_i; i += 1; end }
bench_method("hash", M) { i = 0; while i < M; money.hash; i += 1; end }
bench_method("positive?", M) { i = 0; while i < M; money.positive?; i += 1; end }
bench_method("negative?", M) { i = 0; while i < M; neg.negative?; i += 1; end }
```

</details>

### DIFF
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -10,14 +10,6 @@ class Money
 
   attr_reader :value, :currency
 
-  def zero? = @value.zero?
-  def nonzero? = @value.nonzero?
-  def positive? = @value.positive?
-  def negative? = @value.negative?
-  def to_i = @value.to_i
-  def to_f = @value.to_f
-  def hash = @value.hash
-
   class ReverseOperationProxy
     include Comparable
 
@@ -213,6 +205,14 @@ class Money
     return false unless currency.compatible?(other.currency)
     value == other.value
   end
+
+  def zero? = @value.zero?
+  def nonzero? = @value.nonzero?
+  def positive? = @value.positive?
+  def negative? = @value.negative?
+  def to_i = @value.to_i
+  def to_f = @value.to_f
+  def hash = @value.hash
 
   def coerce(other)
     raise TypeError, "Money can't be coerced into #{other.class}" unless other.is_a?(Numeric)
--- a/lib/money/money.rb
+++ b/lib/money/money.rb
@@ -5,13 +5,18 @@ require 'json'
 
 class Money
   include Comparable
-  extend Forwardable
 
   NULL_CURRENCY = NullCurrency.new.freeze
 
   attr_reader :value, :currency
 
-  def_delegators :@value, :zero?, :nonzero?, :positive?, :negative?, :to_i, :to_f, :hash
+  def zero? = @value.zero?
+  def nonzero? = @value.nonzero?
+  def positive? = @value.positive?
+  def negative? = @value.negative?
+  def to_i = @value.to_i
+  def to_f = @value.to_f
+  def hash = @value.hash
 
   class ReverseOperationProxy
     include Comparable


elfassy (approved):

lib/money/money.rb
```diff
@@ -5,13 +5,18 @@
...
   attr_reader :value, :currency
 
-  def_delegators :@value, :zero?, :nonzero?, :positive?, :negative?, :to_i, :to_f, :hash
+  def zero? = @value.zero?
+  def nonzero? = @value.nonzero?
+  def positive? = @value.positive?
+  def negative? = @value.negative?
+  def to_i = @value.to_i
+  def to_f = @value.to_f
+  def hash = @value.hash
```
elfassy: I suggest moving these methods lower with the other instance methods. Top level is usually reserved for class methods like `def_delegators`

alexcwatt: Good call — moved them down next to `eql?` (which pairs with `hash`), keeping the top of the class for class-level declarations. Done in 1f5932d.

Issue #306, opened 2024-06-20
Allow rounding bypass when creating Money objects


dmasciotra:
## Description

There are some use cases where we would like to have an object contain both the currency and an amount, but not have the library auto round for us based on currency settings. 
An example could be to represent the value of something to a more precise decimal than the currency supports, for purposes of bulk purchases.
A more concrete example could be to set the price of something as $0.057 USD so that when we have 100 of the item, the total cost would be $5.70 USD. 

It should be noted arithmetic and logical operations should use this raw value as well.

elfassy:
proposal: Allow for an option to pass the number of precision to be used

```ruby
Money.decimal_precision = 8

# or as a block

Money.decimal_precision(8) do
  money = Money.new(0.12345678, "USD")
  money.to_s #=> "0.12345678"
  money.format #=> "0.12"
end
```

by default the decimal precision is set by the currency

bdewater:
How about a `decimal_precision` keyword argument that is `nil` by default, which then uses the currency precision. Seems easier to work with than global state, based on the assumption the rounding is often desired and only occasionally not (typically during a series of calculations).

When doing arithmetic, use the highest precision:
```rb
Money.new(1, "USD") + Money.new(0.12345678, "USD", decimal_precision: 8)
# 1.12345678 USD
```

The existing `round` instance method could be changed to default to the currency precision for rounding, instead of the current 0, providing an easy way to get back to a rounded amount before storing in the database or serializing for APIs.

PR #534, opened 2026-09-03
Add explicit decimal precision to Money

derikthiessen-shopify:
Part of https://github.com/Shopify/money/issues/306

### TL;DR

Added opt-in decimal precision to `Money` values, retained additional calculation digits until output boundaries, and preserved the precision contract through arithmetic, allocation, splitting, serialization, and money columns.

### Why Change?

Some prices and thresholds need more decimal places than their currency supports, especially before quantities are applied. Rounding those values during construction loses information, while silently combining or persisting values with incompatible precision can change their meaning.

### What Changed?

- Added a per-value `decimal_precision` option that defaults to the currency's minor units.
- Retained additional digits during calculations and rounded explicitly precise values when rendered, converted to subunits, allocated, split, or persisted through a money column.
- Preserved explicit precision through arithmetic and kept equality, hashing, and ordering value-based.
- Rejected arithmetic and allocation inputs with incompatible declared precision.
- Normalized allocation maxima only when their values are exactly representable at the receiver's precision.
- Preserved raw calculation values through YAML and Active Job transport while JSON uses the rendered value.
- Added fixed `decimal_precision` configuration for money columns so stored values reconstruct without losing their declared precision.
- Documented how to construct, calculate with, and persist subunit-precision values.

Fractional unit prices can retain calculation digits that are not individually payable:

```ruby
unit_price = Money.new("0.0057", "USD", decimal_precision: 3)

unit_price.to_s         # => "0.006"
(unit_price * 100).to_s # => "0.570"
```

Allocation can use the same explicit units while preventing maxima from being rounded into larger limits:

```ruby
amount = Money.new("0.057", "USD", decimal_precision: 3)
amount.allocate_max_amounts(["0.029", "0.028"]).map(&:to_s)
# => ["0.029", "0.028"]
```

Models can declare the precision used for database round trips:

```ruby
class Product < ActiveRecord::Base
  money_column :price, currency_column: :price_currency, decimal_precision: 4
end

product = Product.create!(
  price: Money.new("0.0574", "USD", decimal_precision: 4),
)
product.reload.price.to_s # => "0.0574"
```

### Tests

- [x] Ran the complete local test suite: 781 examples, zero failures, and 100% line coverage.
- [x] Passed RBS validation and Steep type checking.
- [x] Passed RuboCop on all changed production Ruby files.
- [x] Ran Binks and addressed its exact-allocation and subunit-rounding findings.

### Risks

- The per-value `decimal_precision:` API follows the approach proposed in the issue rather than scoped/global configuration. This choice needs confirmation from the Money team before merge.
- Fixed `money_column decimal_precision:` is an explicit persistence contract: writes with a different declared precision are rejected, and explicit-precision writes are rejected when the column has no fixed precision. This policy also needs Money-team confirmation.
- Explicitly precise values may contain raw digits beyond `decimal_precision`; code reading `value` or `to_d` observes those digits, while rendered and discrete-unit output is rounded.


lib/money/money.rb
```diff
@@ -86,32 +86,38 @@ def from_subunits(subunits, currency_iso, format: nil)
 
     def from_json(string)
       hash = JSON.parse(string, symbolize_names: true)
-      Money.new(hash.fetch(:value), hash.fetch(:currency))
+      Money.new(hash.fetch(:value), hash.fetch(:currency), decimal_precision: hash[:decimal_precision])
```
derikthiessen-shopify: will see if we can use `hash.fetch` here instead of the hash accessors to match the existing pattern

lib/money_column/active_record_hooks.rb
```diff
@@ -105,6 +111,14 @@ def validate_currency_compatibility!(column, money, currency_column)
       raise MoneyColumn::CurrencyReadOnlyError, msg
     end
 
+    def validate_decimal_precision_compatibility!(column, money, decimal_precision)
+      return unless money.explicit_decimal_precision?
+      return if money.decimal_precision == decimal_precision
+
+      raise MoneyColumn::PrecisionMismatchError,
+        "Invalid #{column}: Money decimal precision #{money.decimal_precision} does not match money column decimal precision #{decimal_precision.inspect}."
```
derikthiessen-shopify, line 120: so this would reject writes with a different declared precision. I'm not sure if that is the contract that the Money gem maintainers want, so calling it out here

lib/money/money.rb
```diff
@@ -65,17 +65,17 @@ def with_currency(currency, &block)
       with_config(currency: currency, &block)
     end
 
-    def new(value = 0, currency = nil)
-      return new_from_money(value, currency) if value.is_a?(Money)
+    def new(value = 0, currency = nil, decimal_precision: nil)
```
derikthiessen-shopify, line 68: Construction example — explicit precision retains extra calculation digits until an output boundary:

```ruby
unit_price = Money.new("0.0057", "USD", decimal_precision: 3)

unit_price.value.to_s("F") # => "0.0057"
unit_price.to_s               # => "0.006"
```

lib/money/money.rb
```diff
@@ -168,23 +192,25 @@ def <=>(other)
 
   def +(other)
     arithmetic(other) do |money|
-      return self if money.value.zero? && !no_currency?
-      Money.new(value + money.value, calculated_currency(money.currency))
+      result_decimal_precision = calculated_decimal_precision(money)
+      return self if money.value.zero? && !no_currency? && result_decimal_precision == precision_argument
+      Money.new(value + money.value, calculated_currency(money.currency), decimal_precision: result_decimal_precision)
```
derikthiessen-shopify, line 197: Arithmetic example — matching precision propagates through the result, while conflicting declared precision raises `Money::IncompatiblePrecisionError`:

```ruby
price = Money.new("1.000", "USD", decimal_precision: 3)
tax = Money.new("0.057", "USD", decimal_precision: 3)

(price + tax).to_s # => "1.057"
(price * 2).to_s   # => "2.000"
```

lib/money/money.rb
```diff
@@ -281,34 +307,36 @@ def as_json(options = nil)
     if (options.is_a?(Hash) && options[:legacy_format]) || Money::Config.current.legacy_json_format
       to_s
     else
-      { value: to_s(:amount), currency: currency.to_s }
+      hash = { value: to_s(:amount), currency: currency.to_s }
```
derikthiessen-shopify, line 310: JSON is a rendered boundary, so it records the rounded value together with the precision needed to reconstruct its representation:

```ruby
money = Money.new("0.0057", "USD", decimal_precision: 3)
money.as_json
# => { value: "0.006", currency: "USD", decimal_precision: 3 }

Money.from_json(money.to_json).to_s # => "0.006"
```

lib/money/converters/converter.rb
```diff
@@ -5,7 +5,8 @@ module Converters
     class Converter
       def to_subunits(money)
         raise ArgumentError, "money cannot be nil" if money.nil?
-        (money.value * subunit_to_unit(money.currency)).to_i
+        value = money.value.round(money.decimal_precision)
```
derikthiessen-shopify, line 8: Subunit conversion is also an output boundary and rounds instead of truncating retained digits:

```ruby
Money.new("0.0099", "USD", decimal_precision: 2).subunits
# => 1
```

derikthiessen-shopify (commented):
Inline usage examples for the public precision contracts introduced by this PR.

lib/money/allocator.rb
```diff
@@ -114,30 +120,40 @@ def allocate(splits, strategy = nil)
     #     #=> [Money.new(5), Money.new(2)]
     def allocate_max_amounts(maximums)
       allocation_currency = extract_currency(maximums + [__getobj__])
-      maximums = maximums.map { |max| max.to_money(allocation_currency) }
-      maximums_total = maximums.reduce(Money.new(0, allocation_currency), :+)
+      maximums = maximums.map { |max| coerce_maximum(max, allocation_currency) }
```
derikthiessen-shopify, line 123: All maximums are interpreted in the receiver's explicit units and must be exactly representable at that precision:

```ruby
amount = Money.new("0.057", "USD", decimal_precision: 3)
amount.allocate_max_amounts(["0.029", "0.028"]).map(&:to_s)
# => ["0.029", "0.028"]
```

lib/money/splitter.rb
```diff
@@ -15,11 +15,13 @@ def initialize(money, num)
 
     def split
       @split ||= begin
-        subunits = @money.subunits
-        low = Money.from_subunits(subunits / @num, @money.currency)
-        high = Money.from_subunits(low.subunits + 1, @money.currency)
+        units = Helpers.money_to_units(@money)
```
derikthiessen-shopify, line 18: Splitting uses the declared precision as its indivisible unit:

```ruby
Money.new("0.057", "USD", decimal_precision: 3)
  .split(2)
  .map(&:to_s)
# => ["0.029", "0.028"]
```

lib/money/rails/job_argument_serializer.rb
```diff
@@ -4,11 +4,13 @@ class Money
   module Rails
     class JobArgumentSerializer < ::ActiveJob::Serializers::ObjectSerializer
       def serialize(money)
-        super("value" => money.value.to_s("F"), "currency" => money.currency.iso_code)
+        attributes = { "value" => money.value.to_s("F"), "currency" => money.currency.iso_code }
+        attributes["decimal_precision"] = money.decimal_precision if money.explicit_decimal_precision?
```
derikthiessen-shopify, line 8: Active Job transport preserves both the raw calculation value and its declared precision:

```ruby
money = Money.new("0.0574", "USD", decimal_precision: 3)
SomeJob.perform_later(money)
# The job receives value 0.0574 with decimal_precision 3.
```

lib/money_column/active_record_hooks.rb
```diff
@@ -115,14 +130,15 @@ def _assign_attributes(new_attributes)
     module ClassMethods
       attr_reader :money_column_options
 
-      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false, coerce_null: false)
+      def money_column(*columns, currency_column: nil, currency: nil, currency_read_only: false, coerce_null: false, decimal_precision: nil)
```
derikthiessen-shopify, line 133: Model configuration example — the column owns the persistence precision, and explicitly precise assignments must match it:

```ruby
class Product < ActiveRecord::Base
  money_column :price,
    currency_column: :price_currency,
    decimal_precision: 4
end

Product.create!(
  price: Money.new("0.0574", "USD", decimal_precision: 4),
)
```

lib/money/money.rb
```diff
@@ -168,23 +192,25 @@ def <=>(other)
...
+      return self if money.value.zero? && !no_currency? && result_decimal_precision == precision_argument
+      Money.new(value + money.value, calculated_currency(money.currency), decimal_precision: result_decimal_precision)
     end
   end
 
   def -(other)
     arithmetic(other) do |money|
-      return self if money.value.zero? && !no_currency?
-      Money.new(value - money.value, calculated_currency(money.currency))
+      result_decimal_precision = calculated_decimal_precision(money)
```
derikthiessen-shopify: Do not save a database column for the decimal_precision

derikthiessen-shopify: look at the highest precision between the object and the currency and always use that one when we do any arithmetic

derikthiessen-shopify: when converting to string, make sure we are sharing the presentment currency, not the actual value. Keep the decimal precision all throughout all jobs and computations except when showing the values to readers (presentment)

derikthiessen-shopify: incompatible precision error would be removed since we would always use the largest precision for the arithmetic

derikthiessen-shopify: subunits are always based on the currency, not on the precision. There are no decimals in subunits

PR #535, opened 2026-09-13
Use null subunit for non-ISO currencies without a subunit

### LINKED ISSUE #353: Inconsistency in representation when non-ISO currency does not have a subunit
In `currency_iso.yml`, for currencies that do not have a subunit, say, the Japanese yen, there is no `subunit` defined.

However, in `currency_non_iso.yml`, the XFU and GBX currencies define their `subunit` as an empty string.

cpruijsen:
Currencies without a subunit are represented two different ways depending on which file they come
from. `currency_iso.yml` leaves `subunit` unset for them, so it loads as `nil`; `currency_non_iso.yml`
writes `subunit: ''`, which loads as an empty string. Callers that check the subunit therefore have to
handle both, and a truthiness check quietly gets it wrong for one of the two sources.

This drops the empty-string values for `xfu` and `gbx`, the two non-ISO entries with no subunit, so
they load as `nil` like their ISO counterparts. The loader spec gains a case pinning that, alongside
the existing assertion on the same file.

Fixes #353


cpruijsen:
I've signed the CLA.

